PR #55, opened 2015-09-17
examples: Added UHD library dir to CMakeLists.txt

marcusmueller:
without, linkage fails if UHD shared library is not installed in the
default location on Fedora 22.


suttonpd:
Thanks Marcus!

PR #57, opened 2016-04-15
Update FindMATLAB.cmake

krsch:
There are two problems with `readlink`:
1) it might give a path relative to the symlink location,
2) if matlab bin directory is in $PATH it will yield nothing.
The use of `realpath` solves both of them.


waxmihs2902:
thanks

PR #121, opened 2017-11-13
use vector push_back method to allocate required

jgiovatto:
storage for log service objects vs using index operator.


ismagom:
What issue does this fix?

jgiovatto:
Hi,

log_h elements were being added to an empty log_vec via [] operator
resulting in a segfault.
Changed to use vector push_back method, see below...

 std::vector<void*> log_vec;
   for (int i=0;i<args->nof_phy_threads;i++) {
-    log_vec[i] = (void*) log_h;
+   log_vec.push_back((void*)log_h);
   }

Regards
Joe

ismagom:
Has already been fixed in last release. Thanks for spotting it anyway.

PR #138, opened 2018-01-22
Fix bug in random UE ID generation

mrlnc:
UE random ID was limited to `2^40 = 42` instead of `2**40`.

Even with the bitmask corrected, values were limited to `2**31`, because `rand()` returns values up to `RAND_MAX = INT_MAX`, i.e. 31 Bits only.

Therefore bitwise generation is introduced. I'm not sure about `rand()` speeds, so maybe this introduces problems.


mrlnc:
Indeed, also the return value of `rand()` must be expanded to `uint64_t` before shifting. I pushed the new version.

andrepuschmann:
Great! Thanks for that.

PR #139, opened 2018-01-22
Add APN to config and PDN connectivity request

mrlnc:
This allows users to force an APN in PDN connectivity request.

Actually I'm not sure if the following line is ok, since I do not hand a pointer to `nas::init`, but the `struct` itself:
```cpp
srslte_nas_config_t nas_cfg(1, args->apn); /* RB_ID_SRB1 */
nas.init(&usim, &rrc, &gw, &nas_log, nas_cfg); 
```

### DIFF
--- a/lib/include/srslte/common/interfaces_common.h
+++ b/lib/include/srslte/common/interfaces_common.h
@@ -30,6 +30,7 @@
 #include "srslte/common/timers.h"
 #include "srslte/common/security.h"
 #include "srslte/asn1/liblte_rrc.h"
+#include <string>
 
 
 namespace srslte {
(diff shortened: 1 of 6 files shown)


andrepuschmann:
Thanks for that. The way you pass the NAS config is fine. Have you tested that with a MME with two PDNs?

mrlnc:
I have not yet been able to successfully attach with the current release (for other reasons), but had this working on the previous release. Inspected the packets with Wireshark though, looks good to me.

andrepuschmann:
I've just tested here with Amarisoft MME with two PDNs and I get a attach reject. The RRC/NAS messages look ok to me too. What eNB/EPC and config where you using last time?

andrepuschmann:
Thanks.

PR #140, opened 2018-01-26
Configurable MCC/MNC for milenage and xor algorithms

FabianEckermann:
This patch replaces hardcoded MCC and MNC values by config values (otherwise only default values will work).
Added mcc and mnc to hss_args.
(possible duplicate of #137 but testet for milenage)


ismagom:
This is a duplicate of #137

PR #148, opened 2018-02-13
added support for ue/enb/epc to operate without real hardware on a single platform (rev 2)

jgiovatto:
Added faux  rf device (sockrf) to allow a ue and enb to communicate at the rf layer via unix socket.
Streaming samples are carried for src to dst wrapped in a unix socket datagram along with a small
header indicating the seq, gain and sample rate. This allows the ue side to quickly re-sample
as needed for the initial sync sequence. The configuration will specify a device_name = sockrf
and device_args = ue or enb. The rf_sock_recv_with_time() method sets the pace for the message
exchange rate between the enb and ue. For the enb a best effort attempt is made to maintain a
1 msec tti pulse at the "top of the second" and for the ue the action of receiving a msg in a timely fashion is used in an attempt to sync the enb/ue pair. Timing is critical and at a 1msec tti your mileage may vary trying to maintain the beat over an extended period of time. Suggest startup order is epc, ue, then enb with n_prb between 6 and 25. Once the epc is up it can stay up, and the ue/enb pair can be restarted as needed but note that restarting the ue only does not always result in a resync with a previously running enb (TBD). Also when the ue drops the connection it almost never re-establishes the connection with the enb (TBD). If data thruput testing is desired suggest running each application epc/ue/enb in an LXC container to provide network namespace isolation of the instantiated tun devices, although ping -I and tcpdump can crudely be used to exercise the data plane at the tx/rx endpoints respectively.


jgiovatto:
Added "time scaling" to slow down the perceived time of each component and all threads in emulated rf mode only. Attempting to run the enb, ue and epc on the same hardware with a real time rate of 1 msec TTI, has shown to be VERY challenging w/r to kernel scheduling, even when running as root. For most current common desktop type deployment platforms a scale value of 2 (2 msec TTI) set via "cmake -DENABLE_EMURF=ON" sets SRSLTE_TIME_SCALE=2 and allows allows the applications stay in sync much longer. Values of up to 100 or more have still allowed the UE to make a network connection are helpful when running in DEBUG mode where "Observer Effect" complicates the problem you are trying to track.

jgiovatto:
Added "SRSLTE_IPC_CLIENT_SOCK"  and  "SRSLTE_IPC_SERVER_SOCK"  environment variables to allow for running multiple ue/enb pairs on a single platform.
Replace tx_worker threads with forward and store RxQ to alleviate IPC latency, simplifies rx logic and results in slightly better SYNC time.
Some success with sending traffic via 2 enb/ue pairs linked thur a single epc.
Extended time scale to 10x to allow running all processes on a single quad core machine.

jgiovatto:
enb_with_1_ue

jgiovatto:
Replaced unix socket with SV-IPC shared memory  allows for reliable communication between UE/eNB at real frame rates, thus eliminating previous timescale requirement.

Updated dummyrf to working state.

TODO Uplink signal combining.

jgiovatto:
rebase and resubmit

PR #149, opened 2018-02-15
Read MME APN from epc config

FabianEckermann:
This adds an option to read the APN for the EPC MME from the config file


andrepuschmann:
Thanks.

PR #151, opened 2018-02-20
added log format check

jgiovatto:
Changed log service log methods to take string literal to allow for printf format checking and save on string object creation since only a few logs actually passed a string object to the log serivce where these were changed to pass .c_str() in vs adding additional support to handle those few instances. Made a best effort attempt to fix log format warnings where most warnings were somewhat benign vs a few critical cases that had %s with a non string arg or missing all together.


andrepuschmann:
Cherry-picked for the next release. Thanks for that.

PR #163, opened 2018-04-04
Added auth failure and detach to plain nas message

pily112:
Hi,

some more plain NAS message needs to be handled not integrity protected.

Best regards,
David


andrepuschmann:
Thanks David.

PR #166, opened 2018-04-06
Support c++11

jgiovatto:
Minor changes to build for c++11, use of std::isnan, std::isnormal and std::isinf from cmath. 
Should get some good mileage from std::move and auto keyword at a minimum. Tested on Fedora 27 and Ubuntu 16.04.
Init log buffers to null in the event of vasprintf error and use re-entrant version of localtime.


ismagom:
Thanks again for your help and contribution. We will not merge this pull request unfortunately because due to portability reasons we still want to stick to old versions of c++.

PR #172, opened 2018-04-30
NAS key enb derivation fix

pily112:
This branch solves problems wrong key eNodeB derivation, I encountered when the TMSI is not known. See more on the mailing list "RRC Security Mode Command Failure".


andrepuschmann:
Thanks for that David.

PR #173, opened 2018-05-04
fix out of index error

wangqiang1588:
"uint32_t msg_index = sched_info_elem + 1; "  cause msg_index   out of index


andrepuschmann:
Thanks for spotting this. Looks good to me.

ismagom:
Fixes issue #175

PR #188, opened 2018-06-22
changing path to epc.conf while running srsepc

mdasari823:
Need to go to srsepc directory and run the srsepc command otherwise it won't be able to find the path to user_db.csv and throws error.


andrepuschmann:
This will be fixed in new release.

PR #192, opened 2018-07-05
Initial ctxt failure problem

katkoutlia:
I was trying to send an Initial Ctxt Setup Failure to the MME and the eNB was crashing every time I sent the msg. 
The problem was the call of the liblte_s1ap_pack_s1ap_pdu. 
I have done several tests with 2 UEs and is working properly. 

Greetings,
Katerina


andrepuschmann:
Thanks for that. That is already fixed in the next release.

PR #193, opened 2018-07-05
fix comment on libsctp cmake modile

bjoern-r:
The comments had leftover text from mbedtls inside


bjoern-r:
This does not fix the documentation who has was changed in this pull request

PR #198, opened 2018-07-20
Fix crash if args is NULL ,enable multi rx channels setting

wangqiang1588:
Fix crash if args is NULL ,enable multi rx channels setting


andrepuschmann:
Thanks for that.

PR #200, opened 2018-07-25
Change the calculation of UE group A random access preamble selection…

eschreiber-alink:
Change the UE group A preamble selection to choose a random number from [0 nof_groupA_preambles) rather than incrementing on the preambleTransmissionCounter. This helps to reduce collisions in the RRC connection procedure for a multi-UE cell, especially when introducing UEs simultaneously.

Group B preamble selection already uses randomization.


andrepuschmann:
Thanks for that.

PR #207, opened 2018-08-14
make mbms tun device name and UE tun device name configurable

jgiovatto:
mbms device name was hard coded ignoring the optional configuration param mbms_gw.name
this patch removes the hardcoded value "sgi_mb" and applies the configuration value.


jgiovatto:
rebase and resubmit

PR #208, opened 2018-08-14
Daemonize

jgiovatto:
added config option runtime daemonize to ue, enb, epc and mbms
--runtime.daemonize= 0 or 1


jgiovatto:
rebase and resubmit

PR #209, opened 2018-08-14
Ue srand seed

jgiovatto:
Attempt to generate a more unique srand() seed between UE's. Replaced time(NULL) with IMSI and current time microseconds to derive the srand() seed. Previously duplicate contention resolution id's were noticed if each UE was started on the same time in seconds. Had considered using lib_uuid but did not want to introduce and new dependencies. Removed secondary call to srand in rrc assuming that 1 call to srand() is sufficient per process.

Logs below show duplicate random_id with the unlikely but possible duplicate PRACH of 47 resulting in each UE claiming rnti 0x46 and ultimately 172.16.0.2. 

```
ue1.log:15:06:52.639782 [RRC ] [I] Preparing RRC Connection Request using random_id 0x63a69feb9f
ue2.log:15:06:52.639524 [RRC ] [I] Preparing RRC Connection Request using random_id 0x63a69feb9f

enb.log:15:06:53.042872 [PHY0] [I] [05011] PRACH: 0/1, preamble=47, offset=0.0 us, peak2avg=0.0, max_offset=30.0 us

ue1.log:15:06:53.034407 [PHY0] [I] [05006] PRACH: Transmitted preamble=47, CFO=0.00 KHz, nof_sf=1, target_power=-107.0 dBm
ue2.log:15:06:53.034392 [PHY0] [I] [05006] PRACH: Transmitted preamble=47, CFO=0.00 KHz, nof_sf=1, target_power=-107.0 dBm

ue1.log:15:06:53.079419 [PHY0] [I] [05029] set_crnti XXX setting rnti from 0x0 to 0x46
ue2.log:15:06:53.079241 [PHY0] [I] [05029] set_crnti XXX setting rnti from 0x0 to 0x46

ue1.log:15:06:53.308844 [NAS ] [I] Network attach successful. APN: srsapn, IP: 172.16.0.2
ue2.log:15:06:53.308897 [NAS ] [I] Network attach successful. APN: srsapn, IP: 172.16.0.2

```


jgiovatto:
rebase and resubmit

andrepuschmann:
Joe, my apologies again for lacking behind. Looking at the case again it seems like we're using the rand() for the connection request to generate a random UE id. So I don't think we should clutter NAS with the that. Why don't we extent usim_interface_rrc to and let RRC query the IMSI from USIM just before creating the random UE ID?

Issue #180, opened 2018-06-02
bladeRF can't leave BURST mode, so impact next time running


aricwang88:
Actually it's not calibration issue, it's a bug:

=== eNodeB started ===
Type <t> to view trace
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/bladerf.c":356] Enable Module: RX - True
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/bladerf.c":356] Enable Module: TX - True
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/sync.c":326] sync_rx: Worker is idle. Going to reset buf mgmt.
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/sync.c":346] sync_rx: Reset buf_mgmt consumer index
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/sync.c":361] sync_rx: Worker is now running.
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/bladerf.c":356] Enable Module: TX - True
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/bladerf.c":356] Enable Module: TX - True
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/sync.c":849] sync_tx: Worker is now running.
[DEBUG @ "/home/user/bladeRF/host/libraries/libbladeRF/src/sync.c":730] handle_tx_parameters: BURST_START provided while already in a burst.

 TX failed: Invalid operation or parameter (This line printed in rf_blade_send_timed()  function. )

Try to start srsenb many times, if no such error, the bladerf could run normally, if it's have the error even the UE can receive broadcast messages, can not handle uplink data.

aricwang88:
Because rf_blade_send_timed() call twice with  is_start_of_burst = true .

Issue #142, opened 2018-02-02
bladeRF calibration not implemented correctly


ralftoringo:
The current code calls the BladeRF calibration function bladerf_set_correction with 0 for each TX parameter regardless of what is set in the configuration of srsUE or srsENB. I was able to fix this by changing the data type from float to int16_t everywhere (which is what bladerf_set_correction expects).

Furthermore, despite being mentioned in the config files the rx calibration implementation is not finished. It is missing the lines that make it a valid config option in main.cc of both srsUE and srsENB.

However even with these fixes and calibrating the two bladeRFs correctly I am only able to establish a connection between two bladeRFs on very few occasions. SrsUE can find the srsENB signal, but srsENB does not receive anything from srsUE, despite a third SDR showing that data is being sent. The bladeRFs are connected directly to each other with attenuators and they use the same clock.

Connecting srsUE to a commercial base station using bladeRF works very reliably, but connecting a mobile phone to srsENB is also very unreliable.

1. Which SDR do you mainly use for srsENB?
2. It says in the readme that you tested srsLTE with bladeRF. Did you manage to get two of them to connect reliably to each other or use one reliably as base station? If so, do you have further recommendations on how to achieve that?

smarek:
I can confirm, do you have @ralftoringo your attempted patches anywhere public?

@ismagom it's been 2 months since creating this issue, could you please comment at least?

andrepuschmann:
Hey, unfortunately, our bladeRF support is somewhat rusty at this stage, simply because we don't have enough units and hence don't use them for development or testing. If others are willing to take over this job, please go ahead and submit config files, patches, etc. That's why it's open-source.

Kapulara:
We have a couple of units here. Might try to fix this.

aricwang88:
@Kapulara , I'm struggling and suffering from the bladerf calibration issue, both DC and IQ imbalance. I do as the intructions on the bladeRF's wiki, no result.  How to calibrate bladeRF, make it work for srsue? Thanks advance!!!!

aricwang88:
@andrepuschmann , pls gave us a hand.

aricwang88:
The root cause is: start burst tx twice, it cause the internal logic mess, I can rise a pull request to fix this if it possile.  How to get it work: 
1, Comment all the code in xxx_rx_dc_cal(), xxx_tx_dc_cal() as other SDRs.
[...]

PR #210, opened 2018-08-14
Lock pthread mutex at initialization

ecefelix:
The phch_common::init function returns with all mutex unlocked as the
initial state. It allows multiple threads to transmit at the same time
at the begining. It causes BladeRF TX error described in issue #180 and #142.

This patch fixes this bug by locking all the mutex in the init function
except the first mutex.


andrepuschmann:
Thanks for spotting this. The logic has changed in current next and we now use another mechanism for sync so this won't be needed anymore.

PR #212, opened 2018-08-14
Fix SQN incrementation arithmetic

ecefelix:
Current implementation of increment_sqn function increments the 48 bits
SQN directly by 1 which is incorrect according to 3GPP TS 33.102 version
11.5.1 Annex C. This bug causes commercial UEs to report AUTHENTICATION
FAILURE EMM cause #21 "synch failure" when SQN is out of sync.

According to 3GPP TS 33.102 version 11.5.1 Annex C, 48 bits SQN consists
of 43 bits SEQ and 5 bits IND parts where SEQ are incremented during
generation of authentication vector or resync procedure. A
new IND value is also assigned when an authentication vector is
generated. However, resync procedure will use the same IND value.

This patch implements the increment_sqn function according to 3GPP TS
33.102 version 11.5.1 Annex C. A new function increment_seq_after_resync
is added to handle the special case of SQN arithmetic during SQN
resynchronization.

This patch is tested with two comercial UEs (Oneplus One and Oneplus 3T).


alvasMan:
Thanks a lot for this. I had a look at it and it seems fine to me.
The only thing I spotted is that I think that LTE_FDD_ENB_SEQ_HE_MAX_VALUE should be 0x07FFFFFFFFFFUL (2^43-1) instead of 0x7FFFFFFFFFFFUL, considering that the SEQ value has 43 bits.

Can you just confirm with me that I'm not misunderstanding this?

ecefelix:
I think you are right. There is an extra F in 0x7FFFFFFFFFFFUL which makes it (2^47-1). My mistake.
Let me commit a fix.

andrepuschmann:
Thanks guys.

PR #213, opened 2018-08-14
Set AMF to all zeros during resync

ecefelix:
During the resync procedure, the security_milenage_f1_star generates a
MAC_S value that does not match the MAC_S value reported in AUTS. This
posts a security risk.

According to 3GPP TS 33.102 version 11.5.1 section 6.3.3, the AMF used
to calculate MAC-S assumes a dummy value of all zeros so that it does
not need to be transmitted in the clear in the re-synch message.

This patch corrects this problem. MAC_S reported by AUTS matches the
MAC_S calculated using f1_star function after the fix.


andrepuschmann:
Thanks for this.

PR #215, opened 2018-08-17
eNB erab_info clobber and UE integrity check result in duplicate ip address

jgiovatto:
In rapid multiple ue attach condition enb to ue reconfig data could be clobbered before it is sent. Made temporary config storage unique per ue (rnti).
Above condition was causing an integrity check error at the receiver but was ignored resulting in multiple ue's with the same ip address.


jgiovatto:
rebasing

PR #222, opened 2018-08-22
set m1-u multi addr and messaging interface

jgiovatto:
added enb.m1u_bind_addr, enb.m1u_multi_addr and mbms_gw.m1u_bind_addr to allow M1-u over external interface. Best effort approach on allocating new config items to the correct arg_t's and passing config via init api.


jgiovatto:
rebase and resubmit

PR #229, opened 2018-09-10
Fixes for srsLTE on ARM

IgnasJarusevicius:
Fixes for the issues that I encountered while trying to compile and run srsLTE on RPI 3. 

Fix compilation issues with NEON enabled. Also, I noticed that it attempts to compile NEON code even when CMake option 'DISABLE_SIMD' is set (not sure if this is intentional) and I was not able to find a way  to disable NEON code using CMake options (without modifying CMakeList.txt).

Fix runtime timestamp overflow with soapy driver when 'time_t secs' is 32-bit (on 32bit system).


andrepuschmann:
Thanks for that.

PR #230, opened 2018-09-11
fix typo in metrics rssi CMA calculation

igorskh:
Small type when SINR value is used for the RSSI cumulative moving average value instead of the actual RSSI value in the eNB uplink metrics for the PHCH worker.


andrepuschmann:
Thanks for that.

PR #233, opened 2018-09-19
added ability to track tput per bearer

jgiovatto:
added mac_metrics for ul/dl harq retex counts


andrepuschmann:
Thanks for that.

PR #235, opened 2018-09-20
added config option runtime daemonize to ue,enb,epc and mbms (rebase #208)

jgiovatto:
added config option runtime daemonize to ue, enb, epc and mbms
--runtime.daemonize= 0 or 1


andrepuschmann:
Hey Joe, I don't think we'll be merging this PR at this stage so closing it for the public. Thanks for that anyways

PR #236, opened 2018-09-20
added config option for tun dev names and M1 interface rebase(#222 and #207)

jgiovatto:
r#222 added enb.m1u_bind_addr, enb.m1u_multi_addr and mbms_gw.m1u_bind_addr to allow M1-u over external interface. Best effort approach on allocating new config items to the correct arg_t's and passing config via init api.
#207 mbms device name was hard coded ignoring the optional configuration param mbms_gw.name
this patch removes the hardcoded value "sgi_mb" and applies the configuration value.


jgiovatto:
merged

PR #237, opened 2018-09-20
use imsi and current time to generate nas srand seed. (rebase #209)

jgiovatto:
rebase #209


andrepuschmann:
Joe, my apologies again for lacking behind. Looking at the case again it seems like we're using the rand() for the connection request to generate a random UE id. So I don't think we should clutter NAS with the that. Why don't we extent usim_interface_rrc to and let RRC query the IMSI from USIM just before creating the random UE ID?

andrepuschmann:
Thanks for the update. Just wanted to make sure, did you see issue with the same random ID in the connection request or when doing the CHAP auth at NAS?

PR #266, opened 2018-11-19
MME msg encryption

davidrupprecht:
Hi,

I started with the encryption and integrity protection (rework) of NAS message in the MME. Further, I reworked the handling of initial NAS message and transport message to switch case statement. Integrity checking and decryption is done before, the message handling and wrapper for integrity checks and encryption are implemented. 

Best regards,
David


andrepuschmann:
Thanks for that David. We'll have a look at this.

davidrupprecht:
Actually, I did not want to push the last four commits now, but as they are up here, I do not mind. I thought it might be useful to abstract the MME NAS layer from the S1AP interface for testing purposes, that is why I introduced the S1AP interface. The code is not reformatted yet.

PR #267, opened 2018-11-20
add error handling on binding s1c socket

igorskh:
When s1c_bind_addr is set to default and not correct it caused a segmentation fault (noticed with srsepc, and was fine with open air interface EPC) without showing any additional error.

Added an appropriate error message on bind function, the error handling was missing.


andrepuschmann:
Hey Igor, thanks for that. We'll merge that in.

Issue #201, opened 2018-07-26
Support XDG Base Directory specification


SoniEx2:
https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

The XDG Base Directory specification defines base directories for things like config and data files. An example of config files would be enb.conf, epc.conf. An example of data files would be userdb.csv (maybe).

it's nice not having too much clutter in `$HOME`, and being able to easily move most, if not all, config files from one dir to another.

andrepuschmann:
Hey, that looks good. Frankly speaking, we were not aware of that spec. Feel free to submit a PR that implements the spec's suggestions.

andrepuschmann:
As of 18.12 srsLTE now uses ~/.config/srslte as it's default config folder when running as user application.

PR #270, opened 2018-11-30
Make upgrade from Debian official package painless

rubund:
Split in multiple packages to match what is in the official Debian / Ubuntu package.
Add Replaces/Breaks to make upgrade painless
Add systemd service for srsenb, srsepc, srsue
Install default configuration in /etc/srslte
Man pages for srsenb, srsepc, srsue


andrepuschmann:
Hey Ruben, thanks again for looking into the Debian packaging. That effort is much appreciated. I like the idea of splitting the packages. However, I am not so sure about the path you went with the config files. I don't think it's a good idea to have multiple "default" versions of the config files. That becomes difficult to maintain. Why can't we have a script that collects the default ones for the Debian package in the source folders of the individual applications? Also, have you seen issue #201 ? It proposes to use different locations for the config files and we wanted to go that route rather than putting in into etc. Isn't that the default way?

rubund:
Hi Andre,

> Why can't we have a script that collects the default ones for the Debian package in the source folders of the individual applications?

Sure we can. It really depends on the quality of the example configuration files. The default config files I have are just copied from the example files, then tweaked until I have a working configuration here with my SDR. The example files you provide are not usable out of the box when they are placed in /etc/. When it comes to maintenance for the Debian package, there is really not a big difference between the two approaches. In any case, I will have to review and test the changes to the upstream configuration. But if you make a script which prepares the example configuration files such that they work from /etc/, and it works, I am sure I will use it.

> Also, have you seen issue #201 ? It proposes to use different locations for the config files and we wanted to go that route rather than putting in into etc. Isn't that the default way?

There is a clear distinction between user applications and system services. The files in /etc/srslte/* are meant to be used when running as a system service. This is the place for such configuration. Nobody expects configuration to be found under /root/.blablah. It is perfectly fine with configuration under /home/$USER/.blablah when the program is expected to be run as a user. Note that I have also added .service (systemd) which refers to those files in /etc/srslte/*.

This also aligns well with configuration for nextepc (/etc/nextepc/*) or osmo-* (/etc/osmocom/*) which I am also maintaining.

Best regards
Ruben

andrepuschmann:
Hey Ruben,
thanks for the clarification with the XDG specs. Sure, if the default approach for system services is putting configs in /etc/, we can do that. But maybe not every user of srsLTE wants to run the eNB and EPC as services, so we should support both ways. Maybe we can reuse the "cmake/modules/SRSLTE_install_configs.sh.in" script to install in either /etc or ~/.srs (until we support the XDG approach)? Do you think this is feasible? Regarding the configs themselves, yes, the default config should work and you should be able to run a system with those. What exactly did you need to change? I skimmed over them but they look the same.
Thanks
Andre

rubund:
> Maybe we can reuse the "cmake/modules/SRSLTE_install_configs.sh.in" script

Sounds like a good idea.

> What exactly did you need to change? I skimmed over them but they look the same.

Only the paths to the sub-config files ("sib.conf" in enb.conf was changed to "/etc/srslte/sib.conf") had to be changed. If not, it searches for config in /root/.srs/ where I do not have any config files.

> script to install in either /etc or ~/.srs (until we support the XDG approach)? Do you think this is feasible?

The Debian packaging script can call SRSLTE_install_configs.sh.in to install and modify files to /etc/ during package build. No problems with that.

Now there is a postinst script in your Debian package which always installs in /root/.srs. I would almost claim this is always wrong. It is better to motivate the user to call the config-setup-script to get config files to /home/$USER/.src (or .config/..). Or even better, maybe just automatically pick a default config file from /usr/share and put it in the local users configuration folder when starting one of the executables (srsenb, srsue or srsepc). Also print out a message saying that default configuration has been put into the specific directory.

Cheers

rubund:
> Or even better, maybe just automatically pick a default config file from /usr/share and put it in the local users configuration folder when starting one of the executables (srsenb, srsue or srsepc).

I mean something like this (untested):
https://github.com/rubund/srsLTE/commit/e5bc4cf5100e419388c02d91eca2ff08405e83c7

andrepuschmann:
Hey Ruben, sorry for the delay, I am lacking behind this issue. We are currently preparing the next release for srsLTE and would be interested in getting the Debian packing fixed for it. Would you be available offline Github to discuss a few issue on the way to it? If so, please drop me a mail and we discuss next steps.

andrepuschmann:
Just looked at [this](https://github.com/rubund/srsLTE/commit/68038854b6278f97231bbad25ba1f33e9feabf2b). I think the copy should be done in the helper script depending on whether the user wants to install srsLTE as user application or system service. The apps would then check all the default paths (but not do the copying themselves).

akrec:
Hi

SoniEx2:
[XDG Base Directory](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) suggests reading config files from `/etc/xdg/*` (actually `$XDG_CONFIG_DIRS`) when `$HOME/.config/*` isn't available. (technically it doesn't specify how you should handle it except as far as priority is concerned. some applications use the whole chain of configs, whereas others use the first they find. the latter is generally better. the option for config (`--config`?) would skip the system configs.)

I wonder if this would be good enough? @rubund

andrepuschmann:
Most of the changes suggested by @rubund have been integrated into the newest release (18.12). Closing this issue therefore. Please open new issue, if needed, that basis the discussion around the status que in 18.12.

rubund:
> Most of the changes suggested by @rubund have been integrated into the newest release (18.12). 

Great. That should make upgrading from the Debian/Ubuntu official package to the latest srsLTE from your PPA trivial for users.

> XDG Base Directory suggests reading config files from /etc/xdg/* (actually $XDG_CONFIG_DIRS) when $HOME/.config/* isn't available. 

Still, XDG is meant for desktop applications. srsLTE is not a *typical* desktop application. https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard applies for all applications.

PR #274, opened 2018-12-06
Remove undefined increment of an erased iterator.

eschreiber-alink:
The block of code calls an erase on an iterator controlling the for loop, which invalidates the iterator. After the first iteration of the loop, the post-increment on the iterator inside the for statement is not guaranteed to work - on Fedora 29 it causes a segfault. Move the iterator increment operation to a post-increment inside the call to erase.


andrepuschmann:
You're right Eric. Thanks for this.

PR #276, opened 2018-12-10
Fix problems when TA timer is set to infinity

davidrupprecht:
This patch fixes problem that occur when the MAC TA timer is set to infinity.

### DIFF
--- a/srsue/src/mac/demux.cc
+++ b/srsue/src/mac/demux.cc
@@ -245,10 +245,11 @@ bool demux::process_ce(srslte::sch_subh *subh) {
     case srslte::sch_subh::TA_CMD:
       phy_h->set_timeadv(subh->get_ta_cmd());
       Info("Received TA=%d\n", subh->get_ta_cmd());
-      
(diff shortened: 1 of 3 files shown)


andrepuschmann:
Hey David, we finally have our clang-format file ready. It will be included in the new version and we also plan to have all public PRs checked against it. Meanwhile, do you mind reformating your change using the file attached?
[clang-format.zip](https://github.com/srsLTE/srsLTE/files/2663806/clang-format.zip)

davidrupprecht:
Hey, thanks for the clang file. I have reformatted the commit, hope it works well. Do you have any good integration for vscode to format the code automatically before committing?

andrepuschmann:
Cherry-picked on next. Thanks again David. Regarding the auto formating, we don't use vscode but most of us commit on the console anyhow. There are pre-commit hooks like that work awesome. [This] (https://github.com/barisione/clang-format-hooks/) only checks the modified lines for instance. Very useful.

PR #280, opened 2018-12-15
Fixed decoding of multiple s1ap plmns

davidrupprecht:
Fixed decoding of multiple s1ap plmns.


andrepuschmann:
Thanks for that David.

PR #293, opened 2019-01-13
Fallback to AVX2 if AVX512 functions are not support

davidrupprecht:
Hi,

changing the AVX512 errors to warning helps to compile, even if some functions are not AVX512 supported. We now simple downgrade the function to a AVX2 function.

Best regards,
David


davidrupprecht:
Okay, I think that was to fast. Major test cases do not pass, with the fix, I will compile it only with AVX2 support.

PR #298, opened 2019-01-22
Change scheduler_ue get_aggr_level to use current_cfi as the upper limit

eschreiber-alink:
Change scheduler_ue get_aggr_level to use current_cfi as the upper limit for determining the pdcch aggregation level. This prevents the calculation result from exceeding the configured nof_ctrl_symbols.


eschreiber-alink:
Sorry, realizing this is not quite right. Closing.

PR #299, opened 2019-01-24
Prevent ul_dci from being scheduled outside non-mbsfn region on mbsfn subframes.

eschreiber-alink:
Prevent srsenb from crashing when it errantly schedules pdcch carrying ul_dci outside of the non-mbsfn region on mbsfn subframes. Finding this can happen when running with mbsfn enabled, n_prb=15, nof_ctrl_symbols=3 and non_mbsfn_region_length=2 when a UE reports CQI=0. generate_dci is changed to only consider cce locations inside the non-mbsfn area on mbsfn subframes and fail to schedule when pdcch does not fit.

This may not be the cleanest way to do this, but it fixes the problem I'm encountering.


andrepuschmann:
Thanks Eric for raising the issue. The scheduler code has diverged quite a bit since then but we have added the issue to our internal issue tracker and will integrate the restrictions in the next iteration.

PR #308, opened 2019-02-08
bugfix: parse_sibs called before ul_earfcn initialized in enb.cc

legden:
Introduced in commit 4867b16. Function enb::parse_sibs use not yet initialized variable args->rf.ul_earfcn. Some commercial UE can't find network.


andrepuschmann:
Thanks for reporting this. You're right. We've fixed this and it will be included in the next version.

PR #310, opened 2019-02-15
Enb up encryption

davidrupprecht:
Hi, 

I have added control and user plane traffic encryption to the eNodeB based on preferences set in the enb.conf file. For this, I have added / changed the pdcp to rrc interface. Further, I have added a function for preference selection in rrc::ue and we are able to parse the config file for an *ordered* preference list. I would like to add a similar preference selection for the MME encryption / integrity algorithm.

Best regards,
David

### DIFF
--- a/srsenb/src/upper/rrc.cc
+++ b/srsenb/src/upper/rrc.cc
@@ -1914,24 +1914,16 @@ bool rrc::ue::select_security_algorithms() {
 
   bool enc_algo_found = false;
   bool integ_algo_found = false;
-  bool zero_vector = true;
   int i = 0;
-  int j = 0;
   for (i = 0; i < srslte::CIPHERING_ALGORITHM_ID_N_ITEMS; i++) {
     switch (parent->cfg.eea_preference_list[i]) {
     case srslte::CIPHERING_ALGORITHM_ID_EEA0:
       // “all bits equal to 0” – UE supports no other algorithm than EEA0,
-      zero_vector = true;
-      for (j = 0; j < LIBLTE_S1AP_ENCRYPTIONALGORITHMS_BIT_STRING_LEN; j++) {
-        if (security_capabilities.encryptionAlgorithms.buffer[j]) {
-          zero_vector = false;
-        }
-      }
-      if (zero_vector == true) {
-        cipher_algo = srslte::CIPHERING_ALGORITHM_ID_EEA0;
-        enc_algo_found = true;
-        break;
-      }
+      // specification does not cover the case in which EEA0 is supported with other algorithms
+      // just assume that EEA0 is always supported even this can not be explicity signaled by S1AP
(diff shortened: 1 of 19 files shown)


srsenb/src/upper/rrc.cc
```diff
@@ -1881,6 +1901,113 @@ void rrc::ue::send_ue_cap_enquiry()
...
+  bool integ_algo_found = false;
+  bool zero_vector = true;
+  int i = 0;
+  for (i = 0; i < srslte::CIPHERING_ALGORITHM_ID_N_ITEMS; i++) {
+    switch (parent->cfg.eea_preference_list[i]) {
+    case srslte::CIPHERING_ALGORITHM_ID_EEA0:
+      // “all bits equal to 0” – UE supports no other algorithm than EEA0,
+      zero_vector = true;
+      for (int j; j < LIBLTE_S1AP_ENCRYPTIONALGORITHMS_BIT_STRING_LEN; j++) {
+        if (security_capabilities.encryptionAlgorithms.buffer[j]) {
```
andrepuschmann: why is j not initialized in the for loop?

andrepuschmann: If I do set it to zero, there is something going wrong later on and the traffic isn't working

srsenb/src/upper/rrc.cc
```diff
@@ -1881,6 +1901,113 @@ void rrc::ue::send_ue_cap_enquiry()
...
+      break;
+    }
+  }
+
+  for (i = 0; i < srslte::INTEGRITY_ALGORITHM_ID_N_ITEMS; i++) {
+    switch (parent->cfg.eia_preference_list[i]) {
+    case srslte::INTEGRITY_ALGORITHM_ID_EIA0:
+      // “all bits equal to 0” – UE supports no other algorithm than EEA0,
+      zero_vector = true;
+      for (int j; j < LIBLTE_S1AP_INTEGRITYPROTECTIONALGORITHMS_BIT_STRING_LEN;
```
alvasMan: This j also seems uninitialized

andrepuschmann: Yes, setting both to zero breaks it ;-)

alvasMan: Ok, I'm going to investigate. Not initializing seems wrong to me, no sure why setting it to zero is breaking it.

srsenb/src/upper/rrc.cc
alvasMan: OK, so the only reason I can think that not initializing the loop can break stuff is that the uninitialised `j` is always larger than `LIBLTE_S1AP_ENCRYPTIONALGORITHMS_BIT_STRING_LEN`, thus `zero_vector` == true even if it UE supports different algos, thus the eNB will pick EEA0. Not sure why this will break stuff later on though...

andrepuschmann:
Merged in and will be in 19.03. Thanks again @davidrupprecht and @alvasMan

PR #325, opened 2019-04-05
Fix enb putting the SSS in the wrong place with ext CP enabled.

aholtzma:
I noticed that with the extended cyclic prefix enabled, the PSS was missing. It turned out that the SSS had been put on top of the PSS because it was using the wrong cyclic prefix mode.

There is still something wrong with extended prefix mode as phy_dl_test can't pass traffic, but this is a small step towards that working.


andrepuschmann:
Thanks for that.

PR #343, opened 2019-06-12
chap auth random seed

jgiovatto:
Use time in microseconds vs time in seconds to generate a more unique random pattern per ue.
When multiple ue are able to attach simultaneously such as in a emulated environment time in seconds may not be unique per ue.

Removed extra call to srand in srsue process 1 per process should be sufficient.


andrepuschmann:
Thanks for that Joe. We will eventually replace all calls to rand(), and thus srand(), but for the moment that is fine. We merge that in manually into next.

PR #360, opened 2019-08-04
various MAC pcap related fixes

laf0rge:
* MAC pcap generation could be enabled in the config file, but never actually happened -> fix
* MAC pcap files don't include SIB -> update documentation
* WRITE_SIB_PCAP didn't build anymore -> fix

### DIFF
--- a/srsenb/enb.conf.example
+++ b/srsenb/enb.conf.example
@@ -89,6 +89,11 @@ rx_gain = 40
 # add an entry with DLT=147, Payload Protocol=mac-lte-framed.
 # For more information see: https://wiki.wireshark.org/MAC-LTE
 #
+# Please note that this setting will by default only capture MAC
+# frames on dedicated channels, and not SIB.  You have to build with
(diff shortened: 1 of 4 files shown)


mrlnc:
The macro `WRITE_SIB_PCAP` broke before (60bc634386ee2f9f3400d84a22fa0655137119f9) Moving it to ```if (const)``` would make the compiler check the syntax but exclude it from the final code? Is it there for any performance reasons or does it only avoid littering the PCAP? If it's not performance, it would be nice to include in the config.

andrepuschmann:
Thanks for that guys! Just merged the fixing straight into master. I agree that having the SIBs in the PCAP would be great but perhaps only one set of SIBs at the start of the eNB and whenever they change. Putting them always clutters the PCAP and makes it unusable in most (of my) use-cases. Perhaps a config switch like `typedef enum { NEVER, ONCE, ALWAYS } pcap_sib_options_t;`?

PR #367, opened 2019-08-09
Fixed incorrect DCI location assignment in pdsch_enodeb.c example

falkenber9:
Hi all,

the current version of pdsch_enodeb.c example always puts DCI at location ncce=0, L=0, which is incorrect.

Typical UE (and maybe automated tests) won't notice this issue if the signal is clear. DCI will be found anyway if they are decoded at location ncce=0 with L=8 or L=4 (in common search space), since the "empty" CCEs won't negatively influence the bit collection process prior to viterbi decoding.

Swapping two lines of code does fix this issue (cf. pdcch_test.c : 295, setting dci location *after* srslte_dci_msg_pack_pdsch(...)).

Regards
Robert


andrepuschmann:
Merged in PR #371

Issue #368, opened 2019-08-09
Wrong values in the TBS table 


igorskh:
#347  Issue Description ##
Just by chance I have noticed some inconsistency in the tbs table with the 3GPP document in the [lib/src/phy/phch/tbs_tables.h](/srsLTE/srsLTE/tree/master/lib/src/phy/phch/tbs_tables.h) 

## Setup Details ##
Master branch, commit 8c16bcaf30827dc2865f1ffe36fdaa165f92cf95 from 08.08.2019.

General issue, not related to the setup.

## Expected Behavior ##
Exactly the same table as in the [3GPP TS 36.213](https://www.etsi.org/deliver/etsi_ts/136200_136299/136213/12.13.00_60/ts_136213v121300p.pdf) 7.1.7.2.1-1.

## Actual Behaviour ##
Firstly, you can visually check ITBS = 29, Nprb = 12. Expected value = **8760**, got **4760**, p. 66, 3GPP TS 36.213.

https://github.com/srsLTE/srsLTE/blob/8c16bcaf30827dc2865f1ffe36fdaa165f92cf95/lib/src/phy/phch/tbs_tables.h#L233

Then, I've checked the whole table with a script.

Some values are missing in `tbs_table` in the [lib/src/phy/phch/tbs_tables.h](/srsLTE/srsLTE/tree/master/lib/src/phy/phch/tbs_tables.h) (lib/src/phy/phch/tbs_tables.h), some values are wrong.

```
Inconsistent length, i_tbs = 27
Inconsistent length, i_tbs = 28
Incorrect values found, i_tbs = 29
Inconsistent length, i_tbs = 30
Incorrect values found, i_tbs = 31
Incorrect values found, i_tbs = 32
Incorrect values found, i_tbs = 33
```

For incorrect values, see the report attached, they are highlighted with red. First row always shows an expected value, second row an actual value from the srsLTE code.

PDF sample:
[tbs_report.pdf](https://github.com/srsLTE/srsLTE/files/3487248/tbs_report.pdf) 

And the full HTML report:
[report.html.zip](https://github.com/srsLTE/srsLTE/files/3488975/report.html.zip)

## Steps to reproduce the problem ##
Not applicable

## Additional Information ##
I'm just not sure if it's really an issue, or I just don't get the implementation details.

If you confirm the issue, I will generated a table for the C++ code from an ETSI PDF file with a script, that should help to avoid typos, and pull request changes.

[...]

PR #369, opened 2019-08-09
Fix TBS table with auto-generated one (#368)

igorskh:
As described in #368 I submit a pull request with the fixed TBS table, which is automatically generated from the 3GPP TS 36.213 PDF file from ETSI.

Code for table generation is written in Python and I've published it under the MIT License.

You can verify the code here: https://github.com/igorskh/generate_lte_tbs_table


xavierarteaga:
Hi @igorskh, good catch. I will make sure this goes into the next srsLTE release. Thanks.

andrepuschmann:
Thanks for that.

PR #370, opened 2019-08-14
Fix install configs to use the SUDO_GID as well as SUDO_USER.

alex-orange:
This will solve problems where the group is not the same as the user.


andrepuschmann:
applied manually for next version

alex-orange:
To be clear, my submission is bound by the AGPLv3, period.

andrepuschmann:
We've reverted the patch on next.

PR #376, opened 2019-08-26
Fixed inverted handling of fseek result value in srslte_filesource_seek

falkenber9:
Hi all,

I have noticed an inverted handling of ``fseek``'s return value in function ``srslte_filesource_seek(...)``.

A sprious error message is printed, if the filesource is being seeked. Conversely, in case of a seek failure, no message is shown.

The behaviour is fixed in the attached pull request.

Kind Regards
Robert

References: http://man7.org/linux/man-pages/man3/fseek.3.html


andrepuschmann:
Thanks for that Robert.

PR #378, opened 2019-08-28
Fixed duplex spacing for bands 65 to 70

MarkRohrbacher:
The LTE band description array initialized bands >= 65 with the bandwidth instead of its duplex spacing, so the frequency calculation using arfcns above 65536 leads to wrong results.


MarkRohrbacher:
TravisCI seems to be broken, this shouldn't have anything to do with my commit.

```
$ sudo apt-get -qq update
E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/dists/xenial-pgdg/main/binary-amd64/Packages.bz2  Hash Sum mismatch
E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/dists/xenial-pgdg/main/binary-i386/Packages.bz2  Writing more data than expected (209052 > 209044)
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo apt-get -qq update" failed and exited with 100 during .
```

andrepuschmann:
Perhaps a Travis issue. I've restarted the job, let's see if it works this time. Thanks for the fix anyway.

Issue #363, opened 2019-08-06
srsENB doesn't implement UeContextModify procedure


laf0rge:
Our MME gets confused by not seeing any response to S1AP UeContextModifyRequest messages.  It appears the message is parsd in lib/src/asn1/liblte_s1ap.cc, but it is not handled in src/stack/upper/s1ap.cc.

I'll see if I can fix it quickly, but cannot make promises.

Issue #358, opened 2019-08-04
[ENB] No support for CSFB


laf0rge:
I would argue that in every production LTE network today, CSFB (Circuit Switched Fallback) is deployed.  This relates to the fact that many early UEs don't implement VoLTE/IMS, and even those that do often only have explicit white-lists of MCCs/MNCs on which they use VoLTE.  For all other MCC/MNC tuples, CSFB to UTRAN or GERAN is used.

Most of the related work is handled by the MME and the VLR/MSC, see https://osmocom.org/projects/osmomsc/wiki/SGs_Interface for more information.

However, some small bits of this task also need to be performed by the eNB.  There are three different ways how the eNB can send the UE back to GERAN or UTRAN, and they are described in 3GPP TS 23.272 Section 6.3 under 3a/3b/3c.  
* "3a" uses inter-RAT CCO (possibly with NACC) which is rather complex. 
* "3b" uses a simple RRC connection release with rediect to GERAN or UTRAN
* "3c" is an extended version of "3b" as it adds the GSM system information for the target cells to accelerate the process (UE doesn't have to wait for receiving them)

So the simplest approach is "3b".  Supporting it in srsENB would entail not much other than:
* determining that a given RRC connection / UE is doing a CSFB (this can be seen from the InitialContextSetupReq containing the CSFB Indicator)
* including the "redirect" part when sending the RRC Connection Release

I've started to hack on this, and this issue is to keep track of this new feature.

joeldelosangeles:
Hi, is CSFB already supported in srsENB if we will interface to a third-party EPC and GSM core?

joeldelosangeles:
Is the support enough and tested?  Will there be no problems with higher version. (e.g. 20.04 and 20.10)?

## 19.09
  * Add support for CSFB in srsENB

PR #379, opened 2019-08-30
[ENB] Add CFSB support

laf0rge:
This commit introduces CSFB (circuit switched fall-back) capabilities to
srsLTE.  Actually, all the eNB has to do is to send a
RrcConnectionRelease with the RedirectedCarrierInfo IE.

The MME triggers this by the S1AP CS Fallback Indicator IE, which may
be present either in the Initial Context Setup Requst or in the
UE Context Modification Request.

As srsLTE has no support for the UE Context Modification Request at all
yet (!), basic support for this message is introduced in this patch.

Both Mobile Originated and Mobile Terminated CSFB with a coupel of
different UE models have been verified using this patch in a setup
consisting of srsENB attached to an undisclosed EPC connected via SGsAP
to a complete  Osmocom 2G network.

Closes: #358
Closes: #363

### DIFF
--- a/srsenb/src/stack/rrc/rrc.cc
+++ b/srsenb/src/stack/rrc/rrc.cc
@@ -292,9 +292,6 @@ bool rrc::setup_ue_ctxt(uint16_t rnti, LIBLTE_S1AP_MESSAGE_INITIALCONTEXTSETUPRE
     return false;
   }
 
-  if(msg->CSFallbackIndicator_present) {
-    rrc_log->warning("Not handling CSFallbackIndicator\n");
-  }
   if(msg->AdditionalCSFallbackIndicator_present) {
     rrc_log->warning("Not handling AdditionalCSFallbackIndicator\n");
   }
@@ -343,6 +340,14 @@ bool rrc::setup_ue_ctxt(uint16_t rnti, LIBLTE_S1AP_MESSAGE_INITIALCONTEXTSETUPRE
   liblte_pack(msg->SecurityKey.buffer, LIBLTE_S1AP_SECURITYKEY_BIT_STRING_LEN, key);
   users[rnti].set_security_key(key, LIBLTE_S1AP_SECURITYKEY_BIT_STRING_LEN/8);
 
+  // CSFB
+  if (msg->CSFallbackIndicator_present) {
+    if (msg->CSFallbackIndicator.e == LIBLTE_S1AP_CSFALLBACKINDICATOR_CS_FALLBACK_REQUIRED ||
+        msg->CSFallbackIndicator.e == LIBLTE_S1AP_CSFALLBACKINDICATOR_CS_FALLBACK_HIGH_PRIORITY) {
+      users[rnti].is_csfb = true;
+    }
+  }
+
   // Send RRC security mode command
   users[rnti].send_security_mode_command();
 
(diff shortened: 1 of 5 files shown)


srsenb/src/stack/upper/s1ap.cc
```diff
@@ -982,6 +1048,92 @@ bool s1ap::send_initial_ctxt_setup_failure(uint16_t rnti)
...
 }
 
+bool s1ap::send_uectxmodifyresp(uint16_t rnti)
+{
+  if(!mme_connected) {
+    return false;
+  }
+  srslte::unique_byte_buffer_t buf = srslte::allocate_unique_buffer(*pool);
+  if (!buf) {
+    s1ap_log->error("Fatal Error: Couldn't allocate buffer in s1ap::send_initial_ctxt_setup_failure().\n");
```
andrepuschmann: typo in error message (other function name)

srsenb/src/stack/rrc/rrc.cc
```diff
@@ -1563,6 +1642,17 @@ void rrc::ue::send_connection_release()
   dl_dcch_msg.msg.c1().rrc_conn_release().rrc_transaction_id = (uint8_t)((transaction_id++) % 4);
   dl_dcch_msg.msg.c1().rrc_conn_release().crit_exts.set_c1().set_rrc_conn_release_r8();
   dl_dcch_msg.msg.c1().rrc_conn_release().crit_exts.c1().rrc_conn_release_r8().release_cause = release_cause_e::other;
+  if (is_csfb) {
+    rrc_conn_release_r8_ies_s &rel_ies = dl_dcch_msg.msg.c1().rrc_conn_release().crit_exts.c1().rrc_conn_release_r8();
+    rel_ies.redirected_carrier_info_present = true;
+    rel_ies.redirected_carrier_info.set_geran();
+    rel_ies.redirected_carrier_info.geran() = parent->sib7.carrier_freqs_info_list[0].carrier_freqs;
+	    /*
```
andrepuschmann: can this all go away then?

srsenb/src/stack/rrc/rrc.cc
```diff
@@ -354,6 +359,75 @@ bool rrc::setup_ue_ctxt(uint16_t rnti, LIBLTE_S1AP_MESSAGE_INITIALCONTEXTSETUPRE
...
 }
 
+bool rrc::modify_ue_ctxt(uint16_t rnti, LIBLTE_S1AP_MESSAGE_UECONTEXTMODIFICATIONREQUEST_STRUCT *msg)
+{
+  bool err = false;
+  pthread_mutex_lock(&user_mutex);
+
+  rrc_log->info("Modifying context for 0x%x\n", rnti);
+
+  if(users.count(rnti) == 0) {
```
andrepuschmann: here and in a few other lines the clang-format is not applied.

laf0rge: Do you have some kind of guide on how to develop for srsLTE?  I have never heard of "clang-format" before. If there is some kind of automatically enforced coding style, do you have something like pre-commit hooks that one can use so git will automatically perform those?  Or maybe configuration for vim to apply the desired formatting?

laf0rge: ok, I guess I have figured out at least "a" way to use clang-format.  Interestingly, when I wrote the related code (without space between if and the parenthesis) I had to force myself to do it, as I saw the surrounding code was using that.  In Osmocom we use Linux Kernel coding style, and as "if" is not a function, there clearly is a space after it ;)

srsenb/src/stack/upper/s1ap.cc
```diff
@@ -982,6 +1048,92 @@ bool s1ap::send_initial_ctxt_setup_failure(uint16_t rnti)
   return true;
 }
 
+bool s1ap::send_uectxmodifyresp(uint16_t rnti)
+{
+  if(!mme_connected) {
```
andrepuschmann: wrong format

srsenb/src/stack/upper/s1ap.cc
```diff
@@ -982,6 +1048,92 @@ bool s1ap::send_initial_ctxt_setup_failure(uint16_t rnti)
...
+    return false;
+  }
+
+  return true;
+}
+
+
+bool s1ap::send_uectxmodifyfailure(uint16_t rnti, LIBLTE_S1AP_CAUSE_STRUCT *cause)
+{
+  if(!mme_connected) {
```
andrepuschmann: same, and in a few other places below

srsenb/src/stack/rrc/rrc.cc
andrepuschmann: we've introduced a clang-format style file about a year ago I believe and all new code written since then uses this format. However, there is still some "old" code that has, on purpose, not been converted in, e.g., one large clang-format commit. So we will have mixed style until we touch every line at least once more ;-)
We internally all use [this](https://github.com/barisione/clang-format-hooks) commit hook to enforce the formatting. We are working on a script to check this for the public pull requests as well but until this is there we have to manually check this.

srsenb/src/stack/upper/s1ap.cc
laf0rge: I added the script  you referenced as git pre-commit hook, and then did a rebase -i, it didn't  flag anything.  I also manually edited the patch and rebased/edited it a number of times - at no point did the hook flag anything as being incorrect.

In absence of clear guidelines on how actually to check in an automatic/efficient way, I manually looked at the diff and opened each file in vim, navitgated to the sections of code added, marke them in visual mode and then C-K to execute the clang format.

If you insist on suprer-correct formatting in all cases, I strongly recommend that you provide proper tooling and a very clear contribution guide for people to determine if it's correct or not.

laf0rge:
seeing travis-ci being executed here: I'm sure one could also integrate the format check as part of that, giving very clear feedback within seconds after a push.

andrepuschmann:
Yes, this was my intention to integrate in Travis. We are working on dev docs that hopefully will be released soon. Those will include information, among others, regarding  code formatting etc

srsenb/src/stack/upper/s1ap.cc
andrepuschmann: No idea why the pre-commit hook didn't catch the wrongly formatted lines in your case. I think a `rebase -i` should also detect them. In any case, the formatting looks ok now and we can merge the commit. Thanks

northmirko:
I recently conducted CSFB testing with Nokia and Ericsson MME, and it proved to be a challenging task. Achieving this scenario requires the seamless combination of UE's IMSI/EPS attachment. Preservation of P-TMSI in the context (cxtx) is crucial. Specifically, CS service notification needs to be effectively managed within the NAS event loop, and the Extended Service Request should be filled with the appropriate P-TMSI identifier types. While all the necessary programming resources are available, the intricacies of the process make it a demanding undertaking, at least for me. Thank you.

image

PR #381, opened 2019-09-11
Fixes

DreamNik:


andrepuschmann:
Applied to next. Will be part of the new release. Thanks for that.

PR #382, opened 2019-09-17
Fix continuous_tx parsing

litist:
The parsing for continuous_tx was logically inverted.


andrepuschmann:
Thanks Ralf we look into it. Perhaps we fix that internally differently but with the same outcome. Thanks for reporting.

andrepuschmann:
Thanks Ralf. We tested internally and applied to next.

PR #232, opened 2018-09-19
Live MAC Layer Trace in UDP Framing 

davidrupprecht:
Hi,

I have added a live MAC layer trace functionality, that allows to retrieve a live trace on the mac layer. The mac frame is encapsulated into a UDP datagram and can be decoded with wireshark and the heuristic decoder. 

Best regards,
David

PR #383, opened 2019-09-17
Live tracing mac

davidrupprecht:
Rebased on the latest srsLTE version.


davidrupprecht:
Replaces  #232

andrepuschmann:
Hi David, thanks again for preparing the PR. At this stage, we're not going to merge the code into mainline. The main reason for that is that we're planning a bigger refactor of the PCAP functionality that would have a single interface that can be configured to write to file/pipe/socket etc. and perhaps also contains the NR PCAP. If you're interested in working on this as well, we can discuss this further.

PR #385, opened 2019-09-23
Fixed USRP capture sync due to missing channel ptr

davidrupprecht:
Small fix for usrp_capture_sync due to missing channel ptr.


andrepuschmann:
Good catch. Thanks. Applied it to next manually.

PR #389, opened 2019-10-08
Fixed unnecessary arithmetic in srslte_pdcch_common_locations_ncce

omnibusor:


andrepuschmann:
Thanks for that. We'll probably rewrite the code around this and apply the change then.

PR #409, opened 2019-11-14
Always print full RRC messages in UE/eNB

aleksander0m:
Most RRC messages were not being fully printed because the JSON printer does a truly bad job indenting the fields with spaces, so the limit of 1024 characters was reached early. Just ignore the limit and always print full JSON contents.

### DIFF
--- a/srsenb/src/stack/rrc/rrc.cc
+++ b/srsenb/src/stack/rrc/rrc.cc
@@ -166,8 +166,13 @@ void rrc::log_rrc_message(const std::string& source, const direction_t dir, cons
   } else if (rrc_log->get_level() >= srslte::LOG_LEVEL_DEBUG) {
     asn1::json_writer json_writer;
     msg.to_json(json_writer);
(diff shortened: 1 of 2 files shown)


andrepuschmann:
Thanks Aleks for submitting the PR. We've changed this already for the UE in the next upcoming release. But I've cherry-picked `f99af033c2fc7c2c9b2f99258a0c8f194f9539c3` for the eNB.

aleksander0m:
Hey!

> Thanks Aleks for submitting the PR. We've changed this already for the UE in the next upcoming release. But I've cherry-picked `f99af033c2fc7c2c9b2f99258a0c8f194f9539c3` for the eNB.

Is that in a different repo I don't know of? Both my commits were for git master. If I'm not mistaken the UE code had 2 places where RRC messages were being printed: one was already using debug_long() and the other one was using debug(), which is the one I changed.

andrepuschmann:
Yes, we are developing in a private repository and only push that to master every 3 months or so. The next release will be end of December. I've cherry-picked the patch you made for the eNB I think in which we were not using `debug_long` yet. I've manually changed resolved issues while rebasaing. All this will be availbale in 19.12

PR #447, opened 2020-02-01
Support multiple SoapySDR devices using id=# device_arg

jtang613:
This allows multiple SoapySDR devices to be addressed using the id=# device_arg.
This is useful in instances where the device you want to use is not #0, such as running concurrent eNB processes.

### DIFF
--- a/lib/src/phy/rf/rf_soapy_imp.c
+++ b/lib/src/phy/rf/rf_soapy_imp.c
@@ -309,6 +309,7 @@ int rf_soapy_open_multi(char* args, void** h, uint32_t num_requested_channels)
     dev_id = strtol(dev_str, NULL, 0);
     if (dev_id < 0 || dev_id > 10) {
       ERROR("Failed to set device. Using 0 as default.\n");
+      dev_id = 0;
     }
     remove_substring(args, dev_arg);
     remove_substring(args, dev_str);
--- a/lib/src/phy/rf/rf_soapy_imp.c
+++ b/lib/src/phy/rf/rf_soapy_imp.c
@@ -299,9 +299,9 @@ int rf_soapy_open_multi(char* args, void** h, uint32_t num_requested_channels)
   }
 
   // Select Soapy device by id
-  const char dev_arg[]   = "id=";
-  char*      dev_ptr     = strstr(args, dev_arg);
-  int        dev_id      = 0;
+  const char dev_arg[] = "id=";
+  char* dev_ptr = strstr(args, dev_arg);
+  int dev_id = 0;
   if (dev_ptr) {
     char dev_str[64] = {0};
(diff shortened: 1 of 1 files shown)


lib/src/phy/rf/rf_soapy_imp.c
```diff
@@ -298,7 +298,24 @@ int rf_soapy_open_multi(char* args, void** h, uint32_t num_requested_channels)
     printf("\n");
   }
 
-  SoapySDRDevice* sdr = SoapySDRDevice_make(&(soapy_args[0]));
+  // Select Soapy device by id
+  const char dev_arg[]   = "soapy=";
```
andrepuschmann: why don't we just call the param `id=`? seems more intuitive

andrepuschmann: please run clang-format here, the spaces between `=` are too many

jtang613: Fixed.

andrepuschmann:
Can you please check the clang-formating again. See the Travis logs above. It's a minor thing I know, but we have this policy for every PR.
Thanks a lot

```
Checking clang-format between TRAVIS_BRANCH=master and TRAVIS_PULL_REQUEST_BRANCH=soapy_multi_dev
Running code format check between master and  ..
The following code seems to be not formatted correctly:
--- lib/src/phy/rf/rf_soapy_imp.c	(before formatting)
+++ lib/src/phy/rf/rf_soapy_imp.c	(after formatting)
@@ -299,9 +299,9 @@
   }
 
   // Select Soapy device by id
-  const char dev_arg[]   = "id=";
-  char*      dev_ptr     = strstr(args, dev_arg);
-  int        dev_id      = 0;
+  const char dev_arg[] = "id=";
+  char*      dev_ptr   = strstr(args, dev_arg);
+  int        dev_id    = 0;
   if (dev_ptr) {
     char dev_str[64] = {0};
     copy_subdev_string(dev_str, dev_ptr + strnlen(dev_arg, 64));
The command "if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
  # Run only for PRs because target branch is needed to do the clang-format check
  echo "Checking clang-format between TRAVIS_BRANCH=$TRAVIS_BRANCH and TRAVIS_PULL_REQUEST_BRANCH=$TRAVIS_PULL_REQUEST_BRANCH"
  ./run-clang-format-diff.sh "$TRAVIS_BRANCH" "$TRAVIS_PULL_REQUEST_BRANCH"
else
```

jtang613:
Hey, thanks again for the feedback.  Looks like I didn't quite catch what the Travis report was trying to tell me the first time.  Everything is passing now.

PR #450, opened 2020-02-04
github: comment notice on issue template

FFY00:
This way users don't need to remove the text from the issue body, it
will be ignored automatically.

Signed-off-by: Filipe Laíns <lains@archlinux.org>


andrepuschmann:
Cherry-picked on next. Thanks

Issue #453, opened 2020-02-05
Choose CPU extensions manually in the cmake options


FFY00:
Packagers need to able able to choose the target CPU extensions manually as they need to build the project for various different targets. #449 would be the optimal approach, but for now, we should have cmake options to choose the extensions manually.

andrepuschmann:
Fixed in next

PR #454, opened 2020-02-05
cmake: enable choosing CPU extesions

FFY00:
`ENABLE_GENERIC` will disable the resolution of the SSE packages and the
user will have to manually set `HAVE_{AVX,AVX2,SSE}`.

Solves #453

Signed-off-by: Filipe Laíns <lains@archlinux.org>


CMakeLists.txt
```diff
@@ -317,7 +317,10 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   ADD_C_COMPILER_FLAG_IF_AVAILABLE("-Wno-unused-but-set-variable" HAVE_WNO_UNUSED_BUT_SET_VARIABLE)
   ADD_CXX_COMPILER_FLAG_IF_AVAILABLE("-Wno-unused-but-set-variable" HAVE_WNO_UNUSED_BUT_SET_VARIABLE)
 
-  find_package(SSE)
+  if (NOT ENABLE_GENERIC)
```
andrepuschmann, line 320: i see your point but maybe the option should be named differently, like 'AUTO_DETECT_ISA` or so, and put the option into the main CMakeLists.txt with default value set to `true`. Also add the check in the `CMAKE_C_COMPILER` section

FFY00:
Thanks for the feedback. I made the changes.

andrepuschmann:
Thanks Felipe. We cherry-picked that on next.

PR #459, opened 2020-02-20
Fixed typo in invalid ip header length

DavidLin1577:
Note that the word 'lenght' is wrong,
so that 'lenght' should been replaced with 'length'.


andrepuschmann:
Cherry-picked on next. Thanks for that

PR #473, opened 2020-03-19
More generous CSFB

jtang613:
I've observed that srsenb will reject the UEContextModificationRequest produced by my test setup since it includes RegisteredLAI information. That srsenb does not explicitly handle RegisteredLAI is currently logged as a warning, thus it seems appropriate to remove the error indicator and allow the request to proceed.

Test setup:
- srsenb (LTE eNB)
- open5gs (LTE EPC)
- osmocom (GERAN)

Also including updated sib.conf with the necessary sib7.

### DIFF
--- a/srsenb/sib.conf.example
+++ b/srsenb/sib.conf.example
@@ -9,7 +9,7 @@ sib1 =
     (
         {
             si_periodicity = 16;
-            si_mapping_info = [7]; // comma-separated array of SIB-indexes (from 3 to 13). 
+            si_mapping_info = []; // comma-separated array of SIB-indexes (from 3 to 13). 
             // Leave empty or commented to just scheduler sib2
         }
     );
--- a/srsenb/sib.conf.example
+++ b/srsenb/sib.conf.example
@@ -9,7 +9,7 @@ sib1 =
     (
         {
             si_periodicity = 16;
-            si_mapping_info = []; // comma-separated array of SIB-indexes (from 3 to 13). 
+            si_mapping_info = [7]; // comma-separated array of SIB-indexes (from 3 to 13). 
             // Leave empty or commented to just scheduler sib2
         }
     );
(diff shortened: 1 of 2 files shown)


jtang613:
A packet capture is attached.  And a demo of it working is shown at: https://youtu.be/kFEFEk3AK-0
[csfb_s1ap_test_min.zip](https://github.com/srsLTE/srsLTE/files/4364504/csfb_s1ap_test_min.zip)

srsenb/sib.conf.example
```diff
@@ -9,7 +9,7 @@ sib1 =
     (
         {
             si_periodicity = 16;
-            si_mapping_info = []; // comma-separated array of SIB-indexes (from 3 to 13). 
+            si_mapping_info = [7]; // comma-separated array of SIB-indexes (from 3 to 13).
```
andrepuschmann, line 12: could you remove scheduling SIB7 by default, since this is only needed if you run 2G alongside we don't require it by default

jtang613: Fixed - although somehow the commit logged my local user instead of my GitHub account and now the CLA assistant is in a funny state and I'm not sure how to rectify that.

andrepuschmann:
Thanks @jtang613 . Are you happy to sign the CLA for this PR?

andrepuschmann:
> Thanks @jtang613 . Are you happy to sign the CLA for this PR?

Ok, just saw your reply. I guess you can use `git commit --amend` to change author mail again.

jtang613:
Bit of a rocky start there, but I think the commit log is fixed now.

laf0rge:
The SIB7 contents makes sense to me. 

The pull request talks about changes related to RegisteredLAI behavior, but I cannot see any here?

jtang613:
I've removed the `err = true` from srsenb/src/stack/rrc/rrc.cc:408 in to prevent the CSFB from bailing when encountering the RegisteredLAI.

andrepuschmann:
Thanks guys, I'll cherry-pick the commits on next so they'll become available in the next release.

PR #490, opened 2020-05-05
fix parsing of zmq arguments

FabianEckermann:
Use the return value of parse_string function to determine whether a parameter was found.
I experienced issues when the rx_type was set, but the rx_format was not. This led to "Unsupported sample format pub" as the tmp variable is still set to the value of the rx_type.


andrepuschmann:
Thanks Fabian, we've merged that manually and make it available shortly.

PR #496, opened 2020-05-15
[RFC] Fix some minor compile issues with GCC 10.1

greigdp:
GCC 10.1 introduces a few changes that appear to cause minor breakage to existing projects. Nothing too major, but not always immediately obvious from the compiler/linker output.

Three issues encountered when building srsLTE, addressed here:

1. `-fno-common` being enabled by default means that global variables can only be tentatively declared once. If declared without `extern`, then they appear to be multiply declared and compiling fails. [1] Note that srsLTE appears to have two implementations of the relevant code (the edited one below, and the 16 bit AVX2 implementation), so this fix may be a little less straightforward, and this particular definition of `extern int firstGo` may need to go into a header in common by the two implementations [2].

2. Some implicit includes need to be made explicitly now. This looks to be related to [1] where it mentions C++ header dependency issues.

3. A Werror=stringop-truncation warning (treated as error) for strncpy() fails to compile when doing a strncpy from a variable of length X, to a variable of length X, as there is potential on the edge condition to lack space for the tailing null character. In this patch I reduced the permitted length of the incoming parameter, since I imagine 255 vs. 256 character limit will not have any impact.

Refs:

[1] https://gcc.gnu.org/gcc-10/porting_to.html
[2] https://gcc.gnu.org/gcc-10/changes.html

### DIFF
--- a/lib/include/srslte/common/move_callback.h
+++ b/lib/include/srslte/common/move_callback.h
@@ -22,6 +22,9 @@
 #ifndef SRSLTE_MOVE_CALLBACK_H
 #define SRSLTE_MOVE_CALLBACK_H
 
+#include <cstdlib>
+#include <cstdint>
+#include <cstdio>
 #include <cstddef>
 #include <functional>
 #include <type_traits>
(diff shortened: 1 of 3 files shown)


andrepuschmann:
Graig,
thanks for your contribution. We have already fixed compilation with gcc 10.1 in our development tree. Even though we won't pull in this pull-request, we'll leave it open for other users until we release the next version.

suttonpd:
Addressed in f02bfe2cff6bc0685a0418ea61debde3e036de18

PR #518, opened 2020-07-08
tf_packet_filter: Improve support for the different filter types (among which IPv6)

smunaut:


smunaut:
No clue why it would work under GCC and not CLang :/   No warnings during build AFAICT ...

andrepuschmann:
> No clue why it would work under GCC and not CLang :/ No warnings during build AFAICT ...

Have you tried to compile with clang locally and ran the `tft_test`? If so, does it pass on your machine?

smunaut:
Ok, so turns out it was just inconsistent / random ...

The filters also include a mask which wasn't applied before and the tests didn't provide it, so it ended up reading past the buffer, uninitialized data and so sometimes it would pass and sometimes not ...

I now included a commit that updates the test suite with valid filters that include the masks.

srsue/src/stack/upper/tft_packet_filter.cc
```diff
@@ -38,66 +38,147 @@ tft_packet_filter_t::tft_packet_filter_t(uint8_t
   log(log_)
 {
   int idx = 0;
+  uint length_in_bytes;
```
andrepuschmann, line 41: please make the length explicit, i.e. use `uint32_t` and init to `0`. I know it's only used in the IPv6 case but the `switch` is so long it is easily forgotten

andrepuschmann:
Thanks again Sylvain. It looks great. I've rebased the branch on our internal development branch and we go through CI/CD checks there again. I won't merge the PR in the public repo but we will make the changes available in 20.10. Thanks again for your contribution.

andrepuschmann:

PR #523, opened 2020-07-13
Several build fixes

pespin:


andrepuschmann:
Thanks Pau but those should already be fixed in our current `next` branch. Lets leave the PR open as it might be helpful for others with the same issue. I'll close it after we have release 20.10

suttonpd:
Addressed in f02bfe2cff6bc0685a0418ea61debde3e036de18

PR #530, opened 2020-07-30
Fix missing static tag in enb_cfg_parser add_library.

faluco:
add a missing static tag for the enb_cfg_parser  library.


suttonpd:
Fixed with be748fa63d6a4dd991fe4612ed13e54ab583beb8

PR #533, opened 2020-08-14
Add missing device names to enb.conf.example

KloolK:
The enb.conf.example config was missing the options "soapy" and "zmq" for device_name.


andrepuschmann:
Thanks for your contribution. I've cherry-picket the commit, they will shortly be pushed to the next branch in the public repo, which will become srsLTE 20.10.

PR #536, opened 2020-08-21
Circuit-switched Fallback eNodeB Stability

matt9j:
Hello!

This pull request contains two commits addressing issues I encountered while integrating with a system using circuit-switched fallback. The first commit addresses a crash in the RRC message generation code. If the UE sends a circuit switched fallback context modification request, the eNodeB blindly accessed the first element of an asn1 array, which could actually be empty. This commit adds a check to determine if the array is non-empty before attempting to access the first element.

The second commit adds documentation to the sib conf file describing the parameters for sib7. I spent a while looking at the config parser and mapping the parameters to the 3GPP specs to understand what was going on, and want to save future users the same effort : )

Testing:
* debug and release build on archlinux with gcc/g++ version 10.1.0.
* end-to-end circuit switched fallback with one UE to an osmocom 2G network.
* no crash with CSFB in both the case that sib7:explicit_list_of_arfcns is defined and not defined.

Unrelated:
Thanks for putting together this project and releasing it to the wider community! It has allowed me to stay productive at home with a low-power SDR in unlicensed fequencies while I'm unable to access my university lab's equipment. The maturity and stability of the codebase is admirable!

### DIFF
--- a/srsenb/sib.conf.example
+++ b/srsenb/sib.conf.example
@@ -137,6 +137,23 @@ sib3 =
     }
 };
 
+#####################################################################
+# sib7 configuration options (See TS 36.331)
+# Contains GERAN neighbor information for CSFB and inter-rat handover.
+# Must be added to sib1::sched_info::si_mapping_info array parameter to be transmitted
+#
+# t_resel_geran: Cell reselection timer (seconds)
+# carrier_freqs_info_list: A list of carrier frequency groups.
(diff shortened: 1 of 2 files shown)


andrepuschmann:
That looks great, Matt. Thanks a lot for the contribution. I'll cherry-pick the commits. It will be part of the next release.

andrepuschmann:

PR #546, opened 2020-09-13
use SoapySDRDevice_enumerateStrArgs to pass device_args

tyc85:
Hi, 
   In using LimeSDR-USB we have a need to select from a list of available devices. This PR is a simple change to leverage the SoapySDRDevice_enumerateStrArgs API to pass string based args to the SDR driver. Not sure about the original intention to pass NULL for device_args, we hope this can help and does not break anything. The tests we've done are using --rf.device_args to switch between devices of different serial number:
- srsue --rf.device_args=serial=[serial-number-1]
- srsue --rf.device_args=serial=[serial-number-2]

And two applications can be run simultaneously using two different LimeSDR connected to the same host.

### DIFF
--- a/lib/src/phy/rf/rf_soapy_imp.c
+++ b/lib/src/phy/rf/rf_soapy_imp.c
@@ -286,7 +286,8 @@ float rf_soapy_get_rssi(void* h)
 int rf_soapy_open_multi(char* args, void** h, uint32_t num_requested_channels)
 {
   size_t          length;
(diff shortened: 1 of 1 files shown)


andrepuschmann:
Thanks @tyc85 for this contribution. We've found the exact same limitation earlier this summer and have changed the SoapyArgs already. It will be made available next month with the 20.10 release. I'll leave the PR open until then so others have a chance to use the patch in the meantime.

andrepuschmann:

PR #553, opened 2020-09-30
libsrslte_rf: add library version

ciakval:
This patch adds version string to the 'libsrslte_rf' library.
This results in the library being built as
    libsrslte_rf.so.20.04.2
with the following symbolic links
    libsrslte_rf.so.0 -> libsrslte_rf.so.20.04.2
    libsrslte_rf.so -> libsrslte_rf.so.0

The SRSLTE_SOVERSION variable should be increased with each backwards
incompatible change of the library.

Signed-off-by: Jan Remes <jan.remes@invasys.com>


andrepuschmann:
Thanks for your contribution. I've cherry-picket the commit, they will shortly be pushed to the next branch in the public repo, which will become srsLTE 20.10.

PR #568, opened 2020-10-18
Fix "Comparison is always true because ret >= 0" LGTM warnings

djanderson:
Hi, I checked out some LGTM issues as listed here: https://github.com/srsLTE/srsLTE/wiki/Possible-contributions 

There were a few false positives but I did find the "Comparison is always true because ret >= 0" warnings to be valid.

I took the opportunity to lightly refactor and cut out dead code around these warnings. Let me know if you'd prefer me to re-submit with the LGTM fixes only.

-Doug


andrepuschmann:
Hey Doug, thanks for your contribution. I'll cherry-pick and run through our internal CI.

PR #571, opened 2020-10-22
Set same antennas for all Soapy channels

gracid:
This pull request addresses an issue where two or more channels could have different antennas, despite being explicitly set in the configuration file.


andrepuschmann:
That looks good. Thanks for the contribution. I've cherry-picked the commit and applied it manually to the next branch. Will be part of the 20.10 release.

PR #573, opened 2020-10-23
Remove unnecessary Lime calibration step from Soapy implementation

gracid:
When using srsLTE with Lime devices, calibration was performed before any configuration steps have happened, thus making calibration values invalid. Removing Lime specific calibration step from rf_soapy_imp makes so that devices will be automatically calibrated by SoapyLMS on rf_soapy_start_stream call.

Tested and working with srsENB using LimeSDR-USB v1.4 and LimeSDR-Mini v1.2 boards.


andrepuschmann:
My favorite commits. Fixing bugs by removing code. Thanks for the contribution. I've cherry-picked the commit and applied it manually to the next branch. Will be part of the 20.10 release.

PR #575, opened 2020-10-23
eNB: add parameter gtp_ext_addr

mrlnc:
If the eNodeB is behind some gateway, the initial_ctxt_setup_response needs to contain that gateway address rather than the (unreachable) bind address. If unset, the parameter defaults to the bind address.

(#574 was on wrong branch)

### DIFF
--- a/lib/include/srslte/interfaces/enb_interfaces.h
+++ b/lib/include/srslte/interfaces/enb_interfaces.h
@@ -547,6 +547,7 @@ typedef struct {
   uint16_t    mnc;     // BCD-coded with 0xF filler
   std::string mme_addr;
   std::string gtp_bind_addr;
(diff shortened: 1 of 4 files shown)


andrepuschmann:
Thanks for that Merlin. Could you explain perhaps explain the scenario a bit? I assume the gateway you refer to is another machine?

mrlnc:
I had NAT-scenarios in mind like
* srsENB + srsUE with ZMQ in docker containers
* MME outside of docker host.

srsENB would bind to the local docker address (*`gtp_bind_addr`*), but the traffic coming from the core network must first go to the docker host (*`gtp_ext_addr`*).

Chances are that I was just fixing my broken lab setup with this patch 😅 not sure anymore if this is useful to include.

PR #580, opened 2020-10-30
add error msg when uplink frequency is not set

MBCharaf:
I noticed that if the uplink frequency or uplink EARFCN is not set explicitly when the downlink frequency is set explicitly, the uplink frequency is set to the default ul_earfcn without showing any error message.

Added an error message to remind to set the uplink frequency explicitly


MBCharaf:
Not the right error message

PR #581, opened 2020-10-30
Add error message if uplink frequency is not set

MBCharaf:


MBCharaf:
couldn't sign the CLA because of a wrong user name

PR #582, opened 2020-10-30
Add error message if uplink frequency is not set

MBCharaf:
I noticed that if the uplink frequency is not set explicitly when the downlink frequency is set explicitly, the uplink frequency is set to the default ul_earfcn without showing any error message or warning.

Added an error message


andrepuschmann (changes requested):
Hey @MBCharaf . Thanks for the PR. I must admit that the error text and also the section in the code is a bit misleading. Could you perhaps improve/explain it better. Also check the code formatting please. Thanks

PR #620, opened 2021-02-03
Improve reliability of timestamp to tti conversion

FabianEckermann:
Difftime is not reliable as it might use 32 bit calculation, depending on the system. This leads to wrong frame numbers and subframe indices.
I encountered this Issue when testing the pssch_ue exampel on my system and the conversion from timestamp to frame number and sf idx was wrong.


andrepuschmann:
Thanks again for this Fabian. I'll close the PR until the licensing is sorted. It has to be rebased after the new release anyway.

PR #621, opened 2021-02-03
Improve GNSS Sync

FabianEckermann:
Added loop to check for GNSS alignment while syncing.
If the received GNSS signal is weak, synchronization errors might occur while syncing.


andrepuschmann:
Thanks again for this Fabian. I'll close the PR until the licensing is sorted. It has to be rebased after the new release anyway.

PR #632, opened 2021-03-11
Avoid bounded_bitset constructor with size arg as compiler flags buff…

anarkiwi:
…er use uninitialized when optimized.

/root/srsLTE/lib/include/srslte/adt/bounded_bitset.h: In member function 'srsenb::alloc_outcome_t srsenb::sf_grid_t::alloc_ul_data(srsenb::sched_ue*, srsenb::prb_interval, bool)':
/root/srsLTE/lib/include/srslte/adt/bounded_bitset.h:219:32: error: 'newmask.srslte::bounded_bitset<100, true>::buffer[<unknown>]' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  219 |       buffer[i] &= other.buffer[i];
      |                    ~~~~~~~~~~~~^

### DIFF
--- a/srsenb/src/stack/mac/scheduler_grid.cc
+++ b/srsenb/src/stack/mac/scheduler_grid.cc
@@ -482,7 +482,8 @@ alloc_outcome_t sf_grid_t::alloc_ul_data(sched_ue* user, prb_interval alloc, boo
     return alloc_outcome_t::ERROR;
   }
 
-  prbmask_t newmask(ul_mask.size());
+  prbmask_t newmask;
+  newmask.resize(ul_mask.size());
   newmask.fill(alloc.start(), alloc.stop());
   if ((ul_mask & newmask).any()) {
     return alloc_outcome_t::RB_COLLISION;


frankist:
Hey @anarkiwi . Can you share your compiler version? I am compiling with `-Werror` but I do not see this warning.

anarkiwi:
Thanks! So this a tricky issue. We are using this compiler on multiple hosts running current Ubuntu 20.04:

$ g++ --version
g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

The compilation breaks, on a host with this CPU:

model name      : Intel(R) Celeron(R) CPU  J3160  @ 1.60GHz

On other hosts, no problem. It looks very similar to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89733. I speculate that the problem is the IL is not being generated deterministically based on the host CPU. If that's true then that's fairly disturbing. I haven't been able to prove it, though yet.

anarkiwi:
An update. I was able to reproduce this problem in a generic Docker container.

https://github.com/anarkiwi/atomgccpuzzle

If you try to build srsLTE with the supplied Dockerfile (which has qemu run gcc as an atom variant - but just the gcc frontend - the build fails with the error I cited).

frankist:
Thanks a lot @anarkiwi for the detailed analysis. I am going to investigate.

anarkiwi:
I logged a bug also with Ubuntu (it looks like Ubuntu 18 does not have the problem, because of a different gcc version).

https://bugs.launchpad.net/ubuntu/+source/gcc-9/+bug/1918982

nerdasaur:
i'm also seeing the same error as @anarkiwi with same g++ version and Intel Celeron J4105 @ 1.5 GHz

frankist:
Converting the variable `buffer` from C array to `std::array` and initialize it with `{}` seems to fix it for me. I will apply this fix in the next release.

PR #649, opened 2021-04-20
Add feature to provide custom GTP-U (S1) address to be advertised inside S1AP messages

herlesupreeth:
This feature is useful in scenarios where eNB is behind a NAT and EPC on public cloud. Since eNB's GTP-U can only bind to IP address visible in it's host (usually a private IP beind a NAT), it advertises the same to UE in S1AP message resulting in traffic from EPC to eNB (DL traffic) not able to reach eNB. 

This commit solves this issue by providing a means to advertise a different IP address than the one GTP-U UDP socket is bound to. This IP is usually advertised in InitialContextSetupResponse and E-RABSetupResponse S1AP message

In order to make use of this feature, use the following parameter in enb.conf

Example:

`gtp_advertise_addr = 213.61.141.82`


srsenb/src/main.cc
```diff
@@ -72,17 +72,18 @@ void parse_args(all_args_t* args, int argc, char* argv[])
...
-    ("enb.p_a",           bpo::value<float>(&args->enb.p_a)->default_value(0.0f),                  "Power allocation rho_a (-6, -4.77, -3, -1.77, 0, 1, 2, 3)")
+    ("enb.enb_id",             bpo::value<string>(&enb_id)->default_value("0x0"),                       "eNodeB ID")
+    ("enb.name",               bpo::value<string>(&args->stack.s1ap.enb_name)->default_value("srsenb01"), "eNodeB Name")
+    ("enb.mcc",                bpo::value<string>(&mcc)->default_value("001"),                          "Mobile Country Code")
+    ("enb.mnc",                bpo::value<string>(&mnc)->default_value("01"),                           "Mobile Network Code")
+    ("enb.mme_addr",           bpo::value<string>(&args->stack.s1ap.mme_addr)->default_value("127.0.0.1"),"IP address of MME for S1 connection")
+    ("enb.gtp_bind_addr",      bpo::value<string>(&args->stack.s1ap.gtp_bind_addr)->default_value("192.168.3.1"), "Local IP address to bind for GTP connection")
+    ("enb.gtp_advertise_addr", bpo::value<string>(&args->stack.s1ap.gtp_advertise_addr)->default_value("192.168.3.1"), "IP address of eNB to advertise for DL GTP-U Traffic")
```
andrepuschmann, line 81: please make the default value empty here

srsenb/src/stack/upper/s1ap.cc
```diff
@@ -1115,7 +1115,7 @@ bool s1ap::ue::send_initial_ctxt_setup_response(const asn1::s1ap::init_context_s
     auto& item = container.erab_setup_list_ctxt_su_res.value[i].value.erab_setup_item_ctxt_su_res();
     item.transport_layer_address.resize(32);
     uint8_t addr[4];
-    inet_pton(AF_INET, s1ap_ptr->args.gtp_bind_addr.c_str(), addr);
+    inet_pton(AF_INET, s1ap_ptr->args.gtp_advertise_addr.c_str(), addr);
```
andrepuschmann, line 1118: i would propose to set the `gtp_advertise_addr` to `gtp_bind_addr` by default and only overwrite it when the user specifies it

srsenb/src/stack/upper/s1ap.cc
```diff
@@ -1142,7 +1142,7 @@ bool s1ap::ue::send_erab_setup_response(const erab_setup_resp_s& res_)
       auto& item = res.protocol_ies.erab_setup_list_bearer_su_res.value[i].value.erab_setup_item_bearer_su_res();
       item.transport_layer_address.resize(32);
       uint8_t addr[4];
-      inet_pton(AF_INET, s1ap_ptr->args.gtp_bind_addr.c_str(), addr);
+      inet_pton(AF_INET, s1ap_ptr->args.gtp_advertise_addr.c_str(), addr);
```
andrepuschmann, line 1145: same here

andrepuschmann (changes requested):
Hey Supreeth. As always, thanks for your contribution. In order to avoid making changes to existing config files (and our internal testing as well) I'd like to propse to use the normal GTP bind addr by default and only set the custom addr if the user specifies it. Does that make sense?

In addtion to that, and because we are close to the release, would you mind making a PR based on this branch https://github.com/srsLTE/srsLTE/tree/public_lgtm_trial ? This won't be merged directly and I have to cherry-pick it but it makes our lives easier if it was already based on the current dev version.

Thanks

srsenb/src/stack/upper/s1ap.cc
herlesupreeth: done

srsenb/src/stack/upper/s1ap.cc
herlesupreeth: done

herlesupreeth:
@andrepuschmann Sure, makes sense to base it against the public_lgtm_trial branch. Will close this and send PR against that branch

PR #657, opened 2021-05-01
Remove GTPU tunnel when E-RAB is released

herlesupreeth:
Since the GTPU tunnels were not removed upon release of E-RAB,
the max count of tunnels were reached resulting in not able to
have successful E-RAB setup in subsequent attempt


frankist:
I believe that `release_21_04_final` already includes this fix.

PR #587, opened 2020-11-08
Add experimental cuFFT support.

cbalint13:
Enable offload **experimental** FFT processing on cuda based GPUs.

**Goal**

* Proposed very simple patch enable **drop-in** swap of actual main FFTW3 routines, see: [cuFFTW interface](https://docs.nvidia.com/cuda/cufft/index.html#fftw-supported-interface) 
* cuFFT **don't** support R2R so those are skipped in this PR, however they are not used in srsLTE  (**yet**).

**Description**

* To improve the under [4ms](https://youtu.be/uSDKLxwAw2w?t=1067) computation **requirement** in the very tight [timeframe](https://youtu.be/uSDKLxwAw2w?t=1203) only with limited [threads](https://youtu.be/uSDKLxwAw2w?t=1174) is very challenging.

**Evaluation**

 1. According to a simple [benchmark](https://github.com/hurdad/fftw-cufftw-benchmark) things are **slower** on CUDA vs CPU having @1024pts target sizes:
```
Running ./cufftwf-benchmark
---------------------------------------------------------------------------------------
Benchmark                             Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------
cu_fftwf/1024/manual_time         23298 ns        23283 ns        29583 bytes_per_second=335.33M/s items_per_second=43.9524M/s
cu_fftwf/2048/manual_time         27701 ns        27674 ns        25377 bytes_per_second=564.059M/s items_per_second=73.9324M/s
cu_fftwf/524288/manual_time     2002004 ns      1996820 ns          352 bytes_per_second=1.95117G/s items_per_second=261.882M/s
cu_fftwf/1048576/manual_time    4032475 ns      4022351 ns          175 bytes_per_second=1.9374G/s items_per_second=260.033M/s
cu_fftwf/manual_time_BigO          3.85 N          3.84 N    
cu_fftwf/manual_time_RMS              3 %             3 %    
[cbalint@yoda fftw-cufftw-benchmark]$ ./fftw3f-benchmark
2020-11-08T19:12:26+02:00

Running ./fftw3f-benchmark
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
fftwf/1024/manual_time          1644 ns         1669 ns       425523 bytes_per_second=4.63961G/s items_per_second=622.718M/s
[...]

PR #659, opened 2021-05-03
Add downlink PAPR measurement to eNB.

cbalint13:
This PR introduce signal measurements for eNB in terms of quality and dynamics.

* For now it adds PAPR (Peak to Average Power Ratio) measure routine, also available as console command:
```
Available commands: 
          t: starts console trace
          m: downlink signal measurements
          q: quit srsenb
  cell_gain: set relative cell gain
```
* Example (by pressing **m** in CLI):
```
Cell #0 port #0 PAPR = 14.85 dB 
Cell #0 port #1 PAPR = 15.27 dB 
```

Future indicators for the generated signals can be added like e.g.  [CCDF](http://rfmw.em.keysight.com/wireless/helpfiles/n7617/Content/Main/Understanding_CCDF_Curves.htm) (histogram like) to better reflect the PAPR presence.

Measured PAPR for **srseNB is 15dB** on average, regardless PRBs and/or MIMO/SISO setup.

----------
### PAPR implications

Some of PAPR roles: https://www.mpdigest.com/2018/07/24/gans-role-in-5g

* PAPR is a **very important** indicator of radio signal when about getting it on-air with some power.
* [High PAPR](http://www.techplayon.com/papr-peak-average-power-ratio-matters-power-amplifier) will gave impact on final PA **amplifiers design**, implicit on the signals way to the air.

PAPR at **15dB is very high** for practical on-air broadcast:
  * means that to get it to say **1W (30dBm)** a PA of **31W(45dBm)** is needed for **no distorsions**.
  * may stretch the **dynamic range** for most commercial **LNA**, including the ones built in the SDR's rf path.
  * **commercial PA** solutions **for LTE signals** can handle maximum of **9dB** PAPR (e.g. NXP's AirFast LDMOS series)

Various OFDM PAPR limits: https://www.mpdigest.com/wp-content/uploads/2018/07/Figure01.jpg

| OFDM| dB|
|-----------------|---------|
|3G-WCDMA | 3.5dB|
|3G-HSUPA| 6.5dB|
|LTE| 8.5dB|

---------

### Suitable PAPR reduction

A good coverage for PAPR efforts can be found [here](https://docs.google.com/presentation/d/1oO0I4g2Mc7Xy1vqLHsdmc2e200A1l3lFHY7cxXLm9PU/edit?usp=sharing) available on public [github](https://github.com/shahruk10/PAPRnet)

* The classical [Clipping](https://docs.google.com/presentation/d/1oO0I4g2Mc7Xy1vqLHsdmc2e200A1l3lFHY7cxXLm9PU/edit#slide=id.g5653282741_0_1184) is the [simplest](https://docs.google.com/presentation/d/1oO0I4g2Mc7Xy1vqLHsdmc2e200A1l3lFHY7cxXLm9PU/edit#slide=id.g5653282741_1_305) implementation but it degrades the BER. 

* The mentioned [SelectiveMapping](https://docs.google.com/presentation/d/1oO0I4g2Mc7Xy1vqLHsdmc2e200A1l3lFHY7cxXLm9PU/edit#slide=id.g5653282741_1_38) with [template](https://docs.google.com/presentation/d/1oO0I4g2Mc7Xy1vqLHsdmc2e200A1l3lFHY7cxXLm9PU/edit#slide=id.g5653282741_1_313)  fits srseNB, at cost of  random phase elections, zero BER impact.

*  PAPRnet yields state-of-art, at cost of intense tensor-operators, it make sense if srseNB reach to GPU platforms in future.


ismagom:
This looks amazing, thanks for this great contribution! 

I am not very familiar with PAPR reduction techniques, but looking at SelectiveMapping it looks like it requires the receiver to be aware of the changes, same as PAPRnet. Do you think these can be used with COTS UE? 

I wonder if you are aware of any "simple" PAPR technique we could use that is compatible with LTE standard? other than clipping of course

Thanks again!

cbalint13:
> This looks amazing, thanks for this great contribution!
> 
> I am not very familiar with PAPR reduction techniques, but looking at SelectiveMapping it looks like it requires the receiver to be aware of the changes, same as PAPRnet. Do you think these can be used with COTS UE?

Some PAPR reductions like ToneReservation requires receiver side awareness:
* https://wiki.gnuradio.org/index.php/Tone_Reservation_PAPR

SelectiveMapping does not require receiver side awareness:
  * SM "elects" initial phase from random mappings in [1,-1, j, -j ] for a lower PAPR
  * SM finds by "election" an equivalent OFDM version of the same signal.

SM looks a good candidate, would be a way (IMO, easiest so far) to have it in srsLTE (see sample code template).
Not looked thoroughly at way how to add it, might be some catches making it unsuitable (need play into more deeply).

The 9dB PAPR would be an excellent (normal) target for srseNB to achieve.

> 
> I wonder if you are aware of any "simple" PAPR technique we could use that is compatible with LTE standard? other than clipping of course

PAPRnet is fabulous one, but CNN tensor operator are quite intense. I think such approaches would be possible in the future context of #587 (see my latest note there)

> 
> Thanks again!

With pleasure !
~cristian.

cbalint13:
* As update and correction to above info all SM (selective mapping) require some sort of UE awareness:

      a. SM is most effective when re-mapping info is transmitted on side-channel to UE
      b. SM without side-channel but UE should be still aware of OFDM possible permutations

* Other reduction models beyond **clipping** are it's windowed variants (also RTL friendly): [PW-CFR & PC-CFR](https://rfmw.em.keysight.com//wireless/helpfiles/n7614/Content/Main/Crest_Factor_Reduction_Concept.htm)

* The ML based approach like PAPnet have to be trained without any receiver side awareness.

PR #681, opened 2021-06-14
Update srsepc_if_masq.sh

joereith:
This allows bi-directional communications with a COTS UE.


andrepuschmann:
Thanks for your contribution. We'll review this shortly.

rob-zeno:
FYI, I found this PR after I'd debugged and figured out it was missing, but I can confirm that this change fixes the issue for me -- it's exactly what I had to do (wish I'd found it before!).  One thing about this fix -- prior to this change, it was technically possible to run the script before running srsepc.  But now that it references the tunnel device, srsepc needs to be running or the interface won't be found (although I don't know what iptables does in that case).

PR #708, opened 2021-07-28
Minor srsepc proposed fixes 

eschreiber-alink:
When adding a new NAS context to the MME UE ID map, in the check that
the new NAS context matches the context identified by the IMSI, pull from
the IMSI map for the check, not the MME UE ID map.

When resetting the NAS context on an service request integrity check
failure, don't zero the context IMSI. This seems to help with a UE
rejoining the EPC after an extended radio link failure with the ENB.

### DIFF
--- a/srsepc/src/mme/nas.cc
+++ b/srsepc/src/mme/nas.cc
@@ -710,7 +710,7 @@ bool nas::handle_service_request(uint32_t                m_tmsi,
     nas_ctx->reset();
     // But don't erase imsi
     emm_ctx->imsi  = imsi;
-    //emm_ctx->state = EMM_STATE_DEREGISTERED;
+    emm_ctx->state = EMM_STATE_DEREGISTERED;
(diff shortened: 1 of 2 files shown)


srsepc/src/mme/nas.cc
```diff
@@ -708,6 +708,9 @@ bool nas::handle_service_request(uint32_t                m_tmsi,
 
     // Reset and store context with new mme s1ap id
     nas_ctx->reset();
+    // But don't erase imsi
```
andrepuschmann, line 711: can you remove the changes to `nas.cc` from this commit?

andrepuschmann (changes requested):
Thanks for this Eric. Looks good to me. I have only a minor request regarding the commits. It would be great to have the changes to `nas.cc` and `s1ap.cc` in separate commits. Thanks

eschreiber-alink:
No problem - I will resubmit with the changes.

PR #710, opened 2021-07-29
 Minor srsepc proposed fixes - (#708 reworked into 2 commits)

eschreiber-alink:
When adding a new NAS context to the MME UE ID map, in the check that
the new NAS context matches the context identified by the IMSI, pull from
the IMSI map for the check, not the MME UE ID map.

When resetting the NAS context on an service request integrity check
failure, don't zero the context IMSI. This seems to help with a UE
rejoining the EPC after an extended radio link failure with the ENB.


eschreiber-alink:
Closing this pull to conduct further testing in multiple UE networks.

PR #733, opened 2021-10-08
Add pcap over net support for ue

matan1008:
Tested over zmq.
Thanks a lot!


srsue/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@
 # and at http://www.gnu.org/licenses/.
 #
 
+find_package(SCTP REQUIRED)
```
andrepuschmann, line 21: why is this needed for the UE?

srsue/src/CMakeLists.txt
```diff
@@ -43,7 +43,8 @@ target_link_libraries(srsue   ${SRSUE_SOURCES}
                               ${SRSRAN_SOURCES}
                               ${CMAKE_THREAD_LIBS_INIT}
                               ${Boost_LIBRARIES}
-                              ${ATOMIC_LIBS})
+                              ${ATOMIC_LIBS}
+                              ${SCTP_LIBRARIES})
```
andrepuschmann, line 47: this shouldn't be needed for the UE

andrepuschmann (commented):
Thanks for that. Could you check the SCTP thing please.

srsue/CMakeLists.txt
matan1008: I don't really know why it is required (maybe to send the capture over the network?) but without linking with SCTP the build fails with:

```CMake
[ 78%] Linking CXX executable srsue
/usr/bin/ld: ../../lib/src/common/libsrsran_common.a(network_utils.cc.o): in function `srsran::task_details::heap_table_t<srsran::sctp_recvmsg_pdu_task, bool, int>::call(void*, int) const':
network_utils.cc:(.text._ZNK6srsran12task_details12heap_table_tINS_21sctp_recvmsg_pdu_taskEbJiEE4callEPvi[_ZNK6srsran12task_details12heap_table_tINS_21sctp_recvmsg_pdu_taskEbJiEE4callEPvi]+0xe6): undefined reference to `sctp_recvmsg'
collect2: error: ld returned 1 exit status
make[2]: *** [srsue/src/CMakeFiles/srsue.dir/build.make:181: srsue/src/srsue] Error 1
make[1]: *** [CMakeFiles/Makefile2:10845: srsue/src/CMakeFiles/srsue.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
```

andrepuschmann: Mhh, maybe the network utils need to be stripped to have a version without SCTP that can be used in srsUE. I have to say that we are reluctant to do this change now. I'd leave the PR open in case it's useful for others like it is.

PR #750, opened 2021-11-04
emergency handler for NAS PCAP to close file

mrlnc:
This adds emergency handlers for NAS PCAPs to ensure the files are flushed/closed when the application is killed.

Emergency handlers were added with 2c1e9c0c55712601ed7da637d11dade2b09fe3da (thanks @faluco!) to flush logs/PCAPs but NAS is missing.

### DIFF
--- a/lib/src/common/nas_pcap.cc
+++ b/lib/src/common/nas_pcap.cc
@@ -22,9 +22,21 @@
 #include "srsran/common/nas_pcap.h"
 #include "srsran/common/pcap.h"
 #include "srsran/srsran.h"
+#include "srsran/support/emergency_handlers.h"
 #include <stdint.h>
 
 namespace srsran {
+
+/// Try to flush the contents of the pcap class before the application is killed.
+static void emergency_cleanup_handler(void* data)
+{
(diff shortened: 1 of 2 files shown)


lib/src/common/nas_pcap.cc
```diff
@@ -22,10 +22,22 @@
...
 
 namespace srsran {
 
+/// Try to flush the contents of the pcap class before the application is killed.
+static void emergency_cleanup_handler(void* data)
+{
+  reinterpret_cast<nas_pcap*>(data)->close();
+}
+
+nas_pcap::nas_pcap() : enable_write(false), ue_id(0), pcap_file(NULL)
```
mrlnc, line 36: @andrepuschmann I moved variable initialization here, that should be equivalent?

andrepuschmann: it's equivalent. but the current code policy is to init them in the header as well to default values. Would be great if you could follow that. thanks

andrepuschmann (approved):
Excellent. Thanks. Will be included in next release.

PR #763, opened 2021-11-16
srsUE: Add option to filter N_id_1 / SSS to configuration

matan1008:
Hey!
This change will allow forcing the srsUE to select a specific SSS and in my humble opinion will help reducing interference to local operators.
I compiled it and tested it manually (and I ran the existing tests). I found it quite difficult to test this specific feature using the current tests, so any help on that (unless you are fine with merging it without specific tests) would be appreciated.

Thanks for dedicating time!

### DIFF
--- a/srsue/src/main.cc
+++ b/srsue/src/main.cc
@@ -418,6 +418,10 @@ static int parse_args(all_args_t* args, int argc, char* argv[])
      bpo::value<int>(&args->phy.force_N_id_2)->default_value(-1),
      "Force using a specific PSS (set to -1 to allow all PSSs).")
 
+    ("phy.force_N_id_1",
+     bpo::value<int>(&args->phy.force_N_id_1)->default_value(-1),
+     "Force using a specific SSS (set to -1 to allow all SSSs).")
+
     // PHY NR args
     ("phy.nr.store_pdsch_ko",
       bpo::value<bool>(&args->phy.nr_store_pdsch_ko)->default_value(false),
--- a/srsue/src/phy/search.cc
+++ b/srsue/src/phy/search.cc
@@ -55,7 +55,7 @@ search::~search()
   srsran_ue_cellsearch_free(&cs);
 }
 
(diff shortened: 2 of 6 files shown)


srsue/src/phy/search.cc
```diff
@@ -134,6 +135,21 @@ search::ret_code search::run(srsran_cell_t* cell_, std::array<uint8_t, SRSRAN_BC
     Info("SYNC:  Could not find any cell in this frequency");
     return CELL_NOT_FOUND;
   }
+
+  if (force_N_id_1 >= 0 && force_N_id_1 < 168) {
```
andrepuschmann, line 139: use `SRSRAN_NOF_NID_1`

srsue/src/phy/search.cc
```diff
@@ -134,6 +135,21 @@ search::ret_code search::run(srsran_cell_t* cell_, std::array<uint8_t, SRSRAN_BC
...
+    bool N_id_1_found = false;
+    for (uint32_t N_id_2 = 0; N_id_2 < 3; N_id_2++) {
```
andrepuschmann, line 141: use `SRSRAN_NOF_NID_2`

srsue/src/main.cc
```diff
@@ -418,6 +418,10 @@ static int parse_args(all_args_t* args, int argc, char* argv[])
      bpo::value<int>(&args->phy.force_N_id_2)->default_value(-1),
      "Force using a specific PSS (set to -1 to allow all PSSs).")
 
+    ("phy.force_N_id_1",
+     bpo::value<int>(&args->phy.force_N_id_1)->default_value(-1),
```
andrepuschmann, line 422: could you add this to the `ue.conf.example` file? I think also `force_N_id_2` is missing there. So would be good if you could add that too.

andrepuschmann (changes requested):
Thanks for  your contribution. Looks very good to me. Could you address a few comments I've made below. Thanks

andrepuschmann (approved):
Excellent. Thanks for that.

PR #812, opened 2022-02-22
Add additional PLMNs to sib.conf

matan1008:
This change will allow configuring serveral PLMNs (MOCN)

Thanks a lot!

### DIFF
--- a/srsenb/sib.conf.example
+++ b/srsenb/sib.conf.example
@@ -1,3 +1,12 @@
+#####################################################################
+# sib1 configuration options (See TS 36.331)
+#
+# additional_plmns: A list of additional PLMN identities.
+#     mcc: MCC
+#     mnc: MNC
+#     cell_reserved_for_oper:  One of "reserved" or "notReserved", default is "notReserved"
+#
+#####################################################################
 sib1 =
 {
     intra_freq_reselection = "Allowed";
--- a/srsenb/src/enb_cfg_parser.cc
+++ b/srsenb/src/enb_cfg_parser.cc
@@ -79,6 +79,39 @@ bool sib_is_present(const sched_info_list_l& l, sib_type_e sib_num)
   return false;
 }
 
+int field_additional_plmns::parse(libconfig::Setting& root)
+{
+  if (root.getLength() > ASN1_RRC_MAX_PLMN_MINUS1_R14) {
+    ERROR("PLMN-IdentityList cannot have more than %d entries", ASN1_RRC_MAX_PLMN_R11);
+    return SRSRAN_ERROR;
+  }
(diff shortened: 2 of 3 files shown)


andrepuschmann:
Thanks for the PR @matan1008 and sorry for the late response. I am afraid it's too late now for integration in 22.04 but I'll leave the PR open and will include for the next release to have a bit more time for testing. Thanks

srsenb/sib.conf.example
```diff
@@ -1,3 +1,12 @@
+#####################################################################
+# sib1 configuration options (See TS 36.331)
+#
+# additional_plmns: A list of additional PLMN identities.
+#     mcc: MCC
```
andrepuschmann, line 5: how would I specify multiple PLMNs? Can you give an example? Can I select the `reserved` option per PLMN?

matan1008: It would look like:
```
sib1 =
{
    intra_freq_reselection = "Allowed";
    q_rx_lev_min = -65;
    //p_max = 3;
    cell_barred = "NotBarred"
    si_window_length = 20;
    sched_info =
    (
        {
            si_periodicity = 16;
            // comma-separated array of SIB-indexes (from 3 to 13), leave empty or commented to just scheduler sib2
            si_mapping_info = [ 3 ];
        }
    );
    system_info_value_tag = 0;
    additional_plmns = 
    (
        {
            mcc = "345";
            mnc = "67";
        }
    );
};
```
And you can select the `reserved` option by adding `cell_reserved_for_oper = "reserved";`, but the default is `notReserved`

andrepuschmann:
Thanks for this again @matan1008. We'll make sure to include it into the 22.10 release.

andrepuschmann:
Closing PR now, 22.10 will include the changes.

VLiu7:
@matan1008 @andrepuschmann 

Hello, thank you for providing srsran with the `additional_plmns` function. I would like to ask, does this parameter really work? And has this feature been tested?

From the source code, I observed that this `additional_plmns` configuration is eventually saved in the variable `plmn_id_list`, so I searched for it globally.

I observed that `srsue` does not support this function. In `srsue/src/stack/rrc/rrc_cell.cc`, almost all functions read `sib1.cell_access_related_info.plmn_id_list[**0**]`, `additional_plmns` does not work, which I can understand, I can also test with COTS phones.

But in `srsenb` and `srsgnb` I don't see any other `plmn` use either. In `srsgnb/src/stack/rrc/cell_asn1_config.cc`, there is only  `sib1.cell_access_related_info.plmn_id_list[0]` also.  This makes me wonder if the current srsran actually supports the additional_plmns feature. Or it is configured with a completely useless parameter.

PR #814, opened 2022-02-24
using only boost code headers

koh-osug:
The build fails on Ubuntu 20.04-4 when compiling with:

> undefined reference to boost::system::system_category()

The [stack overflow solution](https://stackoverflow.com/questions/9723793/undefined-reference-to-boostsystemsystem-category-when-compiling/50146757#50146757) fixes it.


koh-osug:
I compiled it under an Atom based processor, maybe this is the important context.

PR #869, opened 2022-05-23
Allow sib4 to contain black cell list only.

matan1008:
Hi!
In my humble opinion there was a bug that configuring SIB4 with black cell list only resulted in it being empty.
I tested it with wireshark, I might have misunderstood.

Thanks a lot!


andrepuschmann:
Cherry-picked for next release.

PR #902, opened 2022-06-23
enb,cfg_parser: Fix ac_barring_for_mo_data time typo.

matan1008:
Hey!
I think there is a typo here.
Tested it with wireshark.

Also, I think about adding documentation to sib2 configuration (in a different pr) but it will be pretty long, what do you think?

Thanks a lot!


andrepuschmann:
Cherry-picked. Thanks for that.

PR #923, opened 2022-07-06
Downlink generic nas transport packing

matan1008:
I think that packing generic nas transport message without additional information added a spare byte at the end and the type field of the IEI was discarded (TS 124.301, 8.2.31)


andrepuschmann:
Thanks for this @matan1008

andrepuschmann:
Cherry-picked for next release. Thanks again

PR #948, opened 2022-08-03
Correct f_low_mhz based on 36.104 spec.

joshuazhu78:
B22 fd_low_mhz should be 3510


andrepuschmann:
Thanks for spotting this @joshuazhu78  We pull that in for 22.10. Much appreciated.

PR #963, opened 2022-08-18
Fix for calculation of Ns variable.

EugeneBogush:
I found some incorrect code in paging function.
Variable Ns is calculated wrong According to TS 36.304, Section 7.1

### DIFF
--- a/srsenb/hdr/stack/rrc/rrc_paging.h
+++ b/srsenb/hdr/stack/rrc/rrc_paging.h
@@ -41,7 +41,7 @@ public:
     T(default_paging_cycle_),
     Nb(static_cast<uint32_t>((float)T * nb_)),
     N(std::min(T, Nb)),
-    Ns(std::max(1U, static_cast<uint32_t>(nb_))),
+    Ns(std::max(1U, static_cast<uint32_t>(nb_/T))),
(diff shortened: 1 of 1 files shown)


andrepuschmann:
Thanks for this patch @EugeneBogush. We'll review this shortly.

frankist (commented):
I do not think this is correct. The `nb_` parameter gets initialized as follows:
```
cfg.sibs[1].sib2().rr_cfg_common.pcch_cfg.nb.to_number()
```
whose options are:
```
  static const float options[] = {4.0, 2.0, 1.0, 0.5, 0.25, 0.125, 0.0625, 0.03125};
```
Looking at TS36.304, we have:
```
nB: 4T, 2T, T, T/2, T/4, T/8, T/16, T/32
```
So it seems that `nB` here corresponds to one of the values in `options[]` multipled by `T`, where `T` is the default paging cycle, while `nb_` in the code corresponds to `options[]` values.

Thus, we already have `nb_ == nB / T`. Maybe to improve readibility, I would call `nb_` another name like `nB_x_T`

andrepuschmann:
Thanks for the clarification @frankist. I am closing the PR. @EugeneBogush Please have a look at the spec and let us know if you agree. In case you're still seeing an issue please feel free to reopen and explain.

PR #1018, opened 2022-10-20
Introduce Configurable S1 Connection Timer

jmasterfunk84:
I found myself wanting to reduce the re-try timer between S1 connection attempts.  This makes that timer configurable for those who might be interested in quicker (or slower) retries.


srsenb/src/main.cc
```diff
@@ -268,6 +268,7 @@ void parse_args(all_args_t* args, int argc, char* argv[])
     ("expert.ts1_reloc_overall_timeout", bpo::value<uint32_t>(&args->stack.s1ap.ts1_reloc_overall_timeout)->default_value(10000), "S1AP TS 36.413 TS1RelocOverall Expiry Timeout value in milliseconds.")
     ("expert.rlf_min_ul_snr_estim", bpo::value<int>(&args->stack.mac.rlf_min_ul_snr_estim)->default_value(-2), "SNR threshold in dB below which the eNB is notified with rlf ko.")
     ("expert.max_s1_setup_retries", bpo::value<int32_t>(&args->stack.s1ap.max_s1_setup_retries)->default_value(-1), "Max S1 setup retries")
+    ("expert.s1_connect_timer",  bpo::value<int32_t>(&args->stack.s1ap.s1_connect_timer)->default_value(10), "Connection Retry Timer for S1 connection (seconds)")
```
andrepuschmann, line 271: i think this could be an `uint32_t` to avoid negative values

andrepuschmann (changes requested):
Thanks @jmasterfunk84 for the contribution. I just have a minor comment.

srsenb/src/main.cc
jmasterfunk84: Agreed, and changed!

andrepuschmann:
Thanks again. I've cherry-picked the commit and manually resolved some conflicts created due to other S1 changes. It'll be part of the next release.

PR #1022, opened 2022-10-29
add length check on idx

b1u3s:
Hi,I found ``` LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE``` is 12. So the largest number of idx(also N_emerg_nums) should less than 12.

The define of ``` LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE ``` is here.
```cpp
#define LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE 12
```
It used in:
```cpp
typedef struct {
  LIBLTE_MME_EMERGENCY_NUMBER_STRUCT emerg_num[LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE];
  uint32                             N_emerg_nums;
} LIBLTE_MME_EMERGENCY_NUMBER_LIST_STRUCT;
```
But you haven't check the length of idx, that maybe cause the crash of UE, so I add the length check on idx.

The follow is my test.

in function srsepc/sec/mme/nas.cc/nas::pack_attach_accept,I enable the emergency bumber.
```cpp
attach_accept.emerg_num_list_present = true;
```
and set the attach_accept.emerg_num_list.
```cpp
LIBLTE_MME_EMERGENCY_NUMBER_LIST_STRUCT lists;
LIBLTE_MME_EMERGENCY_NUMBER_STRUCT num;
lists.N_emerg_nums = 14;
attach_accept.emerg_num_list = lists;
```
Then I changed the sent length of emergency number in ``` liblte_mme_pack_emergency_number_list_ie``` function.
```cpp
(*ie_ptr)[0] = 0xff;
```
When srsue received the attach accept msg of MME, The srsUE crash down.
image

### DIFF
--- a/lib/src/asn1/liblte_mme.cc
+++ b/lib/src/asn1/liblte_mme.cc
@@ -3019,9 +3019,13 @@ LIBLTE_ERROR_ENUM liblte_mme_unpack_emergency_number_list_ie(uint8**            
     emerg_num_list->N_emerg_nums = 0;
     while (length < sent_length) {
       idx                                               = emerg_num_list->N_emerg_nums;
+      //add length check on emergency number list
+      if (idx >= LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE) {
+        return (err);
+      }
       emerg_num_list->emerg_num[idx].N_emerg_num_digits = ((*ie_ptr)[length++] - 1) * 2;
       if (emerg_num_list->emerg_num[idx].N_emerg_num_digits > LIBLTE_MME_EMERGENCY_NUMBER_MAX_NUM_DIGITS) {
-        return err;
+        return (err);
       }
 
       emerg_num_list->emerg_num[idx].emerg_service_cat =


lib/src/asn1/liblte_mme.cc
```diff
@@ -3019,6 +3019,9 @@ LIBLTE_ERROR_ENUM liblte_mme_unpack_emergency_number_list_ie(uint8**
     emerg_num_list->N_emerg_nums = 0;
     while (length < sent_length) {
       idx                                               = emerg_num_list->N_emerg_nums;
+      if(idx >= 12) {
```
andrepuschmann, line 3022: no magic numbers, better use `LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE` to check against, also check the code formatting and add a descriptive log/error message

andrepuschmann (changes requested):
Thanks for the contribution. Please see the attached comments.

b1u3s:
> Thanks for the contribution. Please see the attached comments.

OK,I add the error msg in the new changes and use ``` LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE ``` to check the length. Is there anything else I need to improve or change?

lib/src/asn1/liblte_mme.cc
```diff
@@ -3019,6 +3020,10 @@ LIBLTE_ERROR_ENUM liblte_mme_unpack_emergency_number_list_ie(uint8**
     emerg_num_list->N_emerg_nums = 0;
     while (length < sent_length) {
       idx                                               = emerg_num_list->N_emerg_nums;
+      if(idx >= LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE) {
```
andrepuschmann, line 3023: please check the format, a space is missing there, it should be
`if (idx >= LIBLTE_MME_EMERGENCY_NUMBER_LIST_MAX_SIZE) {`

lib/src/asn1/liblte_mme.cc
```diff
@@ -3019,6 +3020,10 @@ LIBLTE_ERROR_ENUM liblte_mme_unpack_emergency_number_list_ie(uint8**
...
+        srsran::console("Error: The length of emergency number list is too long!\n");
```
andrepuschmann, line 3024: I am sorry, checking the code again it seems we are not really handling other unpacking errors with a console print. In order to keep it consistent, just do the return there and that's it.

lib/src/asn1/liblte_mme.cc
```diff
@@ -3019,6 +3020,10 @@ LIBLTE_ERROR_ENUM liblte_mme_unpack_emergency_number_list_ie(uint8**
...
+        return (err);
```
andrepuschmann, line 3025: use `return err;` like everywhere else

andrepuschmann (changes requested):
Thanks again for that. If you could apply the comments and squash all of the commits into one it'd be perfect.

lib/src/asn1/liblte_mme.cc
b1u3s: OK, I'll do like what you say.

lib/src/asn1/liblte_mme.cc
b1u3s: OK, but I found many places use ```(err)`` rather than  ```err``.

andrepuschmann: I am sorry, I just checked the code right below and it was using `return err;`. But as you pointed out the usage in the file is rather mixed. I didn't mean to be a nitpicking person.

b1u3s:
> 

I have applied the  commits into one. To keep the style uniform, I also change the following code:
```cpp
if (emerg_num_list->emerg_num[idx].N_emerg_num_digits > LIBLTE_MME_EMERGENCY_NUMBER_MAX_NUM_DIGITS) {
        return (err);
}
```
I add ``` () ``` around err just like everywhere else.

andrepuschmann (approved):
Thanks for that. I'll cherry-pick it for the next release.

PR #1041, opened 2022-11-20
add length on plmn list

b1u3s:
Add length check on plmn list to avoid memory access errors.


andrepuschmann:
Thanks for this @b1u3s Will be included in next release.

PR #1043, opened 2022-11-22
add escape function to avoid command inject

b1u3s:


andrepuschmann:
Mhh, IMHO the `set_net_admin_caps`  isn't even compiled. I think it can be removed altogether.

PR #1048, opened 2022-12-04
Bugfix/match tac configuration

matan1008:
Hey!
In the current code, configuring TAC as `10` in both the EPC and ENB RR config will yield `Sending S1 Setup Failure - No matching TAC` - this change fixes it.

Thanks!


andrepuschmann:
Thanks for this @matan1008 . Will be included in next release.

PR #1057, opened 2022-12-14
Update main.cc

shotsan:
enabled log saving csv


andrepuschmann:
The parameter is set to `false` by default on purpose. If you want to enable it simply use a modified config file. Thanks

PR #1076, opened 2023-01-05
Add SIB17 parsing

matan1008:
This change will allow configuring SIB17, tested by reading the SIBs with wireshark.

Thanks a lot!

### DIFF
--- a/srsenb/sib.conf.example
+++ b/srsenb/sib.conf.example
@@ -301,3 +301,91 @@ sib7 =
         }
     );
 };
+
+#####################################################################
+# sib17 configuration options (See TS 36.331)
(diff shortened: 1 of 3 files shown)


andrepuschmann:
Thanks for this. But too be honest, I don't see a practical use case for it - we don't do any Wifi offloading. Closing PR, feel free to reopen if you see a use for it.

matan1008:
Out of curiosity - why can't we do offloading?

andrepuschmann:
I am not very familiar with the full mechanism. Is it just adding the SIB and then the UE does all what is needed out of the box? Or is there anything else needed?

matan1008:
I am not familiar with this either, just thought it would be a nice starting point for everyone (including future me) who want to develop this area

PR #1078, opened 2023-01-05
enb,parser: Fix options formatting

matan1008:
Before the change, setting report_amount to 10 printed:
```
Invalid option: 
 for enum field "report_amount"
Valid options:  , , ,, ,  , @, �
Error parsing field cell_list in section cell_list
```

Now it prints:
```
Invalid option: 
 for enum field "report_amount"
Valid options:  1, 2, 4, 8, 16, 32, 64, -1
Error parsing field cell_list in section cell_list
```


andrepuschmann:
Thanks @matan1008 , will be included in next release.

Issue #802, opened 2022-02-07
srs support for adalmpluto with soapysdr


fpgasdr:
## Issue Description ##
Hi, I know the support for the analog devices is limited due timestamp not implemented with analog devices drivers but from  other issue I know there is some test than can be made with this devices like cellsearch(it's working in my case and cell are detected) and PDSCH PSSCH examples but the last ones are not working due sample rate error.

## Setup Details ##
I have ubuntu 20 with the last srsRAN version installed from source and soapysdr witha soapysdrpluto installed and try to run the pssch_ue and pdsch_enodeb example  from build/lib/examples folder:

## Actual Behaviour ##
the output from running both example are the same:

Soapy has found device #0: device=plutosdr, driver=plutosdr, uri=usb:3.4.5, 
Soapy has found device #1: device=plutosdr, driver=plutosdr, uri=ip:pluto.local, 
Selecting Soapy device: 0
Setting up Rx stream with 1 channel(s)
[INFO] Using format CF32.
[INFO] Auto setting Buffer Size: 32768
[INFO] Set MTU Size: 32768
Setting up Tx stream with 1 channel(s)
[INFO] Using format CF32.
[INFO] Has direct TX copy: 1
[INFO] Auto setting Buffer Size: 32768
[INFO] Set MTU Size: 32768
Available device sensors: 
 - xadc_temp0
 - xadc_voltage0
 - xadc_voltage1
 - xadc_voltage2
 - xadc_voltage3
 - xadc_voltage4
 - xadc_voltage5
 - xadc_voltage6
 - xadc_voltage7
 - xadc_voltage8
 - adm1177_current0
 - adm1177_voltage0
 - ad9361-phy_temp0
 - ad9361-phy_voltage2
Available sensors for Rx channel 0: 
State of gain elements for Rx channel 0 (AGC supported):
 - PGA: 71.00 dB
State of gain elements for Tx channel 0 (AGC not supported):
 - PGA: 80.00 dB
Rx antenna set to A_BALANCED
Tx antenna set to A
RF device 'soapy' successfully opened
Setting sampling rate 5.76 MHz
/home/julian/Documents/SDR/srsRAN/lib/examples/pdsch_enodeb.c:796: Could not set sampling rate

the sample rate could not be set. how should I deal with this?.
what could be done with my setup "out of the box"?.

andrepuschmann:
[...]

PR #1082, opened 2023-01-12
Fix for Pluto AD9363 sample rate delta

cloudwiser:
As per reopened issue [#802](https://github.com/srsran/srsRAN/issues/802) and thread with @andrepuschmann

### DIFF
--- a/lib/examples/pdsch_ue.c
+++ b/lib/examples/pdsch_ue.c
@@ -484,7 +484,7 @@ int main(int argc, char** argv)
     signal(SIGINT, sig_int_handler);
 
     /* set receiver frequency */
-    printf("Tunning receiver to %.3f MHz\n", (prog_args.rf_freq + prog_args.file_offset_freq) / 1000000);
+    printf("Tuning receiver to %.3f MHz\n", (prog_args.rf_freq + prog_args.file_offset_freq) / 1000000);
     srsran_rf_set_rx_freq(&rf, prog_args.rf_nof_rx_ant, prog_args.rf_freq + prog_args.file_offset_freq);
 
     uint32_t ntrial = 0;
@@ -495,7 +495,7 @@ int main(int argc, char** argv)
         ERROR("Error searching for cell");
         exit(-1);
       } else if (ret == 0 && !go_exit) {
-        printf("Cell not found after %d trials. Trying again (Press Ctrl+C to exit)\n", ntrial++);
+        printf("Cell not found after [%4d] attempts. Trying again... (Ctrl+C to exit)\n", ntrial++);
       }
     } while (ret == 0 && !go_exit);
 
@@ -504,13 +504,14 @@ int main(int argc, char** argv)
       exit(0);
     }
 
+    #define MAX_SRATE_DELTA 2 // Hz allowable delta between requested and actual sample rate
     /* set sampling frequency */
     int srate = srsran_sampling_freq_hz(cell.nof_prb);
     if (srate != -1) {
       printf("Setting sampling rate %.2f MHz\n", (float)srate / 1000000);
-      float srate_rf = srsran_rf_set_rx_srate(&rf, (double)srate);
-      if (srate_rf != srate) {
-        ERROR("Could not set sampling rate");
+      float srate_rf = srsran_rf_set_rx_srate(&rf, (double)srate);     
+      if (abs(srate - srate_rf) > MAX_SRATE_DELTA) {
+        ERROR("Could not set sampling rate : wanted %d got %f", srate, srate_rf);
         exit(-1);
       }
     } else {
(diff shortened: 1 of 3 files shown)


lib/examples/npdsch_enodeb.c
```diff
@@ -180,7 +186,7 @@ void parse_args(int argc, char** argv)
 
   if (!output_file_name && rf_freq == 0) {
     usage(argv[0]);
-    printf("\nError! Either RF frequency or output filename need to be specified.\n");
+    printf("\nError : either RF frequency or output filename needs to be specified\n");
```
andrepuschmann, line 189: nitpick, but we'd prefer `Error: either ...`, so no space behind error

lib/examples/npdsch_enodeb.c
```diff
@@ -197,12 +203,12 @@ void base_init()
   // init memory
   sf_buffer = srsran_vec_cf_malloc(sf_n_re);
   if (!sf_buffer) {
-    perror("malloc");
+    perror("Error : malloc for sf_buffer");
```
andrepuschmann, line 206: same goes for all such cases

lib/examples/npdsch_enodeb.c
```diff
@@ -218,13 +224,19 @@ void base_init()
...
-    if (srsran_rf_open(&radio, rf_args)) {
-      fprintf(stderr, "Error opening rf\n");
-      exit(-1);
+    if (strlen(rf_dev) > 0) {
+      if (srsran_rf_open_devname(&radio, rf_dev, rf_args, NOF_TX_ANT)) {
+        fprintf(stderr, "Error opening RF device %s\n", rf_dev);
+        exit(-1);
+      }
     }
+    else
```
andrepuschmann, line 233: formatting seems wrong here. Can you use clang-format with the provide file (in the repository root), please?

lib/examples/pdsch_ue.c
```diff
@@ -504,13 +504,14 @@ int main(int argc, char** argv)
       exit(0);
     }
 
+    #define MAX_SRATE_DELTA 2 // Hz allowable delta between requested and actual sample rate
```
andrepuschmann, line 507: I'd move the define also up in to where all other macros are defined

lib/examples/pssch_ue.c
```diff
@@ -263,13 +263,15 @@ int main(int argc, char** argv)
     printf("Set RX freq: %.6f MHz\n",
            srsran_rf_set_rx_freq(&radio, prog_args.nof_rx_antennas, prog_args.rf_freq) / 1e6);
     printf("Set RX gain: %.1f dB\n", prog_args.rf_gain);
-    int srate = srsran_sampling_freq_hz(cell_sl.nof_prb);
 
+    #define MAX_SRATE_DELTA 2 // Hz allowable delta between requested and actual sample rate
```
andrepuschmann, line 267: same here

lib/examples/pssch_ue.c
```diff
@@ -263,13 +263,15 @@ int main(int argc, char** argv)
...
+    /* set sampling frequency */
+    int srate = srsran_sampling_freq_hz(cell_sl.nof_prb);
     if (srate != -1) {
-      printf("Setting sampling rate %.2f MHz\n", (float)srate / 1000000);
+      printf("Setting sampling rate %.2f MHz\n", (float)srate / 1000000);
```
andrepuschmann, line 271: note the white spaces at the end. I'd suggest to setup clang-format in your preferred code editor. Unfortunately, we don't have strict format checks in the public CI

andrepuschmann (changes requested):
Hey Nick, thanks a lot for your contributions. I do have a few minor comments regarding those. Let me know what you think. Thanks

lib/examples/npdsch_enodeb.c
cloudwiser: No problem Andre - will amend

lib/examples/npdsch_enodeb.c
cloudwiser: OK - will ensure there is consistency as per the above

lib/examples/pdsch_ue.c
cloudwiser: OK - I didn't like it here eithe... but didn't know whether it should be promoted into a `.h` file...or just placed at the head of the `pdsch_ue.c`. I'll go with the latter unless you advise otherwise

lib/examples/pssch_ue.c
cloudwiser: Ack - as per above

lib/examples/pssch_ue.c
cloudwiser: Got it - will do

cloudwiser:
@andrepuschmann Hi Andre - updates as per your review requests above. Thanks.

andrepuschmann:
> @andrepuschmann Hi Andre - updates as per your review requests above. Thanks.

Looks good now. Thanks for that.

xavierarteaga (approved):
It looks fine to me too :ok_hand:

cloudwiser:
Thanks both!

Issue #1247, opened 2023-11-21
highly intermittent attach with srsenb and srsue with bladerf 2.0 micro


nitinvjain:
## Issue Description ##
I am referring to :
Document reference: https://buildmedia.readthedocs.org/media/pdf/srslte-docs/latest/srslte-docs.pdf

I am using srsRAN_4G along with bladeRF 2.0 micro Xa4 in a cabled setup as per following connections:
gNodeB Tx1 (40 dB tx_gain --> UE Rx1 (Agc)gnodeB Rx1 (Agc) <-- UE Tx1 (40 dB tx_gain)No GPSDO used for bladeRF.
Basically, I am using srsenb and srsue on 2 different machine.srsenb and srsepc are running on same machine.
Main issue is that attach is too intermittemt and 1 out of 10 times only, it happens. That too sometimes, it works without any release uptp 20-25 mins and sometimes, it releases within 2-3 mins itself.

I tried to play with time_adv_nsamples in UE side by changing it in steps of 5 right from 27 (default) to 140 but no change in behaviour. I am sharing all configurations by attching zip file having conf files as well as .log files.Please point out to me any obvious issues.My understanding based on logs is that DL is probably working fine and UL is quite flaky.Please share any insights to move forward.

I have also referred issues as per following details:
https://github.com/srsran/srsRAN_4G/issues/688

https://github.com/srsran/srsRAN_4G/issues/1105 
## Setup Details ##
Accordingly, I have changed default change in srate in source code and in ue.conf (srate=11.52 e6)
Other Configs:
bladeRF related:bladeRF  2.0 micro XA4bladeRF> version
  bladeRF-cli version:        1.8.0-git-unknown  libbladeRF version:         2.4.1-git-unknown
  Firmware version:           2.4.0-git-a3d5c55f  FPGA version:               0.14.0 (configured by USB host)
Ubuntu versions: Distributor ID: UbuntuDescription: Ubuntu 22.04.3 LTSRelease: 22.04Codename: jammy
srsRAN_4G:
git logcommit fa56836b14dc6ad7ce0c3484a1944ebe2cdbe63b (HEAD -> master, tag: release_23_04_1, origin/master, origin/HEAD)Merge: 921f17484 737c5ae71Author: Codebot <[codebot at softwareradiosystems.com](https://lists.srsran.com/mailman/listinfo/srsran-users)>Date:   Mon Jun 19 15:03:18 2023 +0200
    Merge branch 'next' into agpl_next
[...]

PR #1275, opened 2024-01-23
Nj blade rf 2311 changes

nitinvjain:
Hi, 
Changes in this forked branch take care of issue reported in #1247 

https://github.com/srsran/srsRAN_4G/issues/1247

Please see if you want to merge main changes to master branch.

Let me know if I need to help in that.

Regards

### DIFF
--- a/configs/enb.conf
+++ b/configs/enb.conf
@@ -160,7 +160,7 @@ filename = /tmp/enb.log
 file_max_size = -1
 
 [gui]
-enable = false 
+enable = true 
 
 #####################################################################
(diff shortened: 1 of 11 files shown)


andrepuschmann:
Thanks for the investigation and for providing the PR. However, I am not sure that the srsRAN repo is the best place to apply those fixes. I'd rather see the root cause fixed in the respective driver, not working around it in the application that just uses the provided API.
That being said, however, it's probably a good idea to have the fixes here so interested folks can import them.

PR #1392, opened 2024-09-24
Fix: correct count value for NAS decryption

liuxiaoxinxinxin:


andrepuschmann:
What's the issue here, what has been fixed? Is there a description and test?

XGallagher15:
## Summary

Fix wrong parameter passed to the NAS deciphering functions. The argument to `security_128_eea1/2/3()` should be `ctxt_base.rx_count`.

## Changes

- Pass `ctxt_base.rx_count` in the following branches:
  - `CIPHERING_ALGORITHM_ID_128_EEA1`
  - `CIPHERING_ALGORITHM_ID_128_EEA2`
  - `CIPHERING_ALGORITHM_ID_128_EEA3`

```c
case CIPHERING_ALGORITHM_ID_128_EEA2:
  security_128_eea2(&ctxt_base.k_nas_enc[16],
                    count_est,
                    ctxt_base.rx_count,
                    bearer_id,
                    SECURITY_DIRECTION_DOWNLINK,
                    &pdu->msg[seq_offset + 1],
                    ...
```

## Testing

- [x] Interoperability test with **free5GC**: NAS security procedure and ciphered NAS messages OK
- [x] Interoperability test with **Open5GS**: NAS security procedure and ciphered NAS messages OK

@andrepuschmann Hi, This is the description and test of this PR. Thanks!

alvasMan:
@liuxiaoxinxinxin `count.rx_base` does not take as a parameter the actual message received, which strikes me as odd. 

Can you elaborate why should it be using `count.rx_base` instead of the count estimation?

XGallagher15:
@alvasMan Hi, Thanks for review the code.

The short answer is that `ctxt_base.rx_count` **does** reflect the actually-received message; the SQN just gets folded in one step earlier, so it's the right value to use here. `cipher_decrypt()` is only ever called after a successful `integrity_check()` (`nas.cc:485-487`, and `nas_5g.cc:198-202` for the 5G path). Inside `integrity_check()` we compute `count_est = (ctxt_base.rx_count & 0x00FFFF00u) | pdu->msg[seq_offset]` and, on success, write it back with `ctxt_base.rx_count = count_est` (`nas_base.cc:92,127-130`, per TS 24.301 §4.4.3.3). So by the time we decrypt, `ctxt_base.rx_count` already equals the count derived from *this* message's SQN — and since a single NAS message must use the **same NAS COUNT** for both integrity protection and ciphering, using `ctxt_base.rx_count` guarantees decryption uses exactly the value that was just validated. The local `count_est` we replaced was actually the bug: it computed the count with a hardcoded offset, `(ctxt_base.rx_count & 0x00FFFF00u) | pdu->msg[5]`, using index `5` instead of `seq_offset`. `nas_base` is shared between LTE (`LTE_SEQ_OFFSET = 5`) and 5G (`SEQ_5G_OFFSET = 6`), so on LTE it happened to be correct, but on the 5G path — which is exactly what's exercised against free5GC/Open5GS — `pdu->msg[5]` reads the last MAC byte instead of the sequence number, producing a wrong count → wrong keystream → garbled decryption. Switching to `ctxt_base.rx_count` sidesteps that hardcoded-offset bug and keeps decrypt consistent with the integrity check. One minor follow-up: the leftover `count_est` at `nas_base.cc:189` is now only used in the debug log but still carries the `pdu->msg[5]` issue — harmless, but changing `[5]` → `[seq_offset]` would keep the logged "Received count" from being misleading on 5G.

alvasMan:
OK, so the real problems are then:

1. The cipher_decrypt is estimating the count again, even though it does not need it (because the integrity check already updated RX_COUNT).
2. The estimation in the cipher_decript is wrong, as it does not use the seq_offset.

Can you then add the following to your MR:

* Fully remove the count estimation in the cipher_decrypt function.
* Update the logging message to something along the lines of 
```
    logger.debug("Decrypting PDU. count=%d", ctxt_base.rx_count);
```
* Add a comment at the top of the function saying something like:
```
/// The RX_COUNT state variable has been already updated after checking integrity,
/// as such we do not need to estimate it again.
```

alvasMan (approved):
Awesome, thank you for this :)

liuxiaoxinxinxin:
@alvasMan 

Thanks for the review!

andrepuschmann:
Thanks @liuxiaoxinxinxin . I've cherry-picked the changes to our internal repo first for some additional CI tests. They'll get released to the public shortly.

andrepuschmann:
@liuxiaoxinxinxin Just to confirm. Are you also the author of https://github.com/srsran/srsRAN_4G/pull/1550? I see different Git accounts.

liuxiaoxinxinxin:
@andrepuschmann yes, I am also the author of #1550 and #1549

PR #1463, opened 2025-03-24
 srsenb: implement `expert.sctp_nodelay`

fixeria:
It may be desirable to disable the [Nagle's algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) in some environments by setting `expert.sctp_nodelay` to `true`. For instance, Erlang/OTP seems to have problems with handling incoming SCTP packets containing multiple DATA chunks.


pespin:
IMHO it probably makes sense to actually enable NODELAY by default on S1AP in a follow-up PR.

PR #1542, opened 2026-08-19
Fix typo in enb command line options

philippuhl:
There is a stray ")" in the command line option `expert.sctp_max_init_timeo)` (should be `expert.sctp_max_init_timeo` instead)

This PR removes it.


andrepuschmann:
Thanks @philippuhl - we'll include that in the next release

PR #1543, opened 2026-08-25
gtpu: reject PDUs shorter than the GTP-U header

CaffinatedSquirrel-OG:
`gtpu_read_header()` subtracts the base or extended header length from `pdu->N_bytes` without first checking that the PDU is at least that long. `N_bytes` is a `uint32_t`, so a datagram shorter than the header wraps it to a value near 2^32 and the function still returns `true`. Every consumer downstream then treats that as the payload length.

On the eNB this reaches PDCP, where `undelivered_sdus_queue::add_sdu()` uses it as a `memcpy` length into a pooled buffer. ASAN reports `memcpy-param-overlap`; an unsanitised build segfaults. The S1-U socket is unauthenticated, so the datagram that triggers it needs no credentials.

The 5G stack already has both checks - `srsRAN_Project` does this in `gtpu_dissect_pdu()` at `lib/gtpu/gtpu_pdu.cpp:215` and `:259`. This adds the same two guards here, plus a bound on the `add_sdu()` `memcpy` as defence in depth.

Well-formed PDUs are unaffected: the base and extended header paths parse byte-for-byte as before.

### Background

This came from a private report to `security@srs.io` on 2026-06-03. Andre asked whether I'd prepare a fix and contribute it upstream, so here it is.

Two notes on scope, so nobody reads more into this than it says:

- The **EPC/S-GW** side of that original report is closed by mutual agreement. srsEPC was never a commercial product, and my own TUN probe on 2026-08-18 showed the kernel reads nothing there (errno 105). This PR is the **eNB** path only, which the dev-status page still lists as Under Maintenance.
- I have **not** claimed code execution anywhere. What's demonstrated is a crash, scored `AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`.

Happy to retarget this at a different branch or split the PDCP hunk out if you'd rather keep the parser fix and the defence-in-depth bound separate - just say which you prefer.


andrepuschmann:
Thanks for this @CaffinatedSquirrel-OG . A fix has been developed meanwhile and is now merged into master. Closing this PR therefore.

CaffinatedSquirrel-OG:
No problem at all, Andre — makes sense to take it through your own tree. Glad it's sorted, and thanks for the credit line in the commits.

PR #1544, opened 2026-09-03
cmake: detect 64-bit ARM by explicit processor names

AndreiGosman:
`CMAKE_SYSTEM_PROCESSOR` is tested with `MATCHES`, which is a substring match, and the two ARM guards in `CMakeLists.txt` disagree about what 64-bit ARM looks like.

### What goes wrong

**`HAVE_NEONv8` is never set on a platform that reports `arm64`.** The guard tests `MATCHES "aarch64"`, so such a build gets `HAVE_NEON` without `HAVE_NEONv8` and silently takes the pre-ARMv8 code paths in `lib/src/phy/modem/demod_soft.c` and `lib/src/common/arch_select.cc`. Nothing fails; it just uses the wrong implementation.

**32-bit options reach 64-bit compilers.** `-mfloat-abi=hard -mfpu=neon` are guarded by `MATCHES "arm"`, which is true for `arm64` as a substring. Where the compiler rejects them the build stops before any source file is reached:

```
clang++: error: unsupported option '-mfloat-abi=' for target 'arm64-apple-darwin25.5.0'
```

The asymmetry is why this is easy to miss: `aarch64` does not contain `arm`, so a Linux ARM64 build trips the second problem but not the first, and neither shows up on x86.

### The change

Name the 64-bit processor strings explicitly in both guards.

### Effect

Evaluated the old and new conditions against the processor strings CMake produces:

| processor | `HAVE_NEONv8` | 32-bit flags |
|---|---|---|
| `arm` | no → no | yes → yes |
| `armv7l` | no → no | yes → yes |
| `armv7` | no → no | yes → yes |
| `aarch64` | yes → yes | no → no |
| `arm64` | no → **yes** | **yes** → no |

Only `arm64` changes, and it now behaves as `aarch64` does.

### Context

Found while building on Darwin, which reports `arm64`. The `-mfloat-abi` half is specific to compilers that reject those options, but the `HAVE_NEONv8` half affects any system using that spelling regardless of compiler.


andrepuschmann:
Thanks for this. Cherry-picked to internal branch for testing on Rpi3 arm32

andrepuschmann:
Thanks - the patch has been cherry-picked for internal testing and will make it into the next release.

PR #1545, opened 2026-09-04
cell_search: fix broken getopt parsing (arguments after the first are silently ignored)

AndreiGosman:
### Reproduction

```
cell_search -b 3 -g 50 -a type=b200 -s 1250 -e 1260
```

Only `-b 3` takes effect. The gain, the device arguments and both EARFCN bounds are discarded, so the scan runs at the default gain, with the default device arguments, across the whole band. There is no error and no warning, and the scan still produces output, which makes this easy to mistake for a hardware or propagation problem rather than arguments that never arrived.

### Cause

Two things combine in `parse_args`.

The option string declares no arguments:

```c
while ((opt = getopt(argc, argv, "agsendvb")) != -1) {
```

Without the colons, `getopt` treats every option as a flag and stops at the first one followed by a value.

The switch then reads the value from `argv[optind]` rather than `optarg`:

```c
case 'b':
  band = (int)strtol(argv[optind], NULL, 10);
```

That hides the first problem for exactly one option. For the first flag, `optind` happens to point at the value because nothing has consumed it, so the value is picked up and the command appears to work. From the second flag onwards nothing lines up.

### Fix

Declare the arguments and read them from `optarg`, which is what the switch was already written for.

```c
while ((opt = getopt(argc, argv, "a:b:d:s:e:n:g:v")) != -1) {
```

Every case in the switch already expects a value except `-v`, so the option string now matches the code.

### Notes

Found while scanning for LTE cells with a B200 series device, where a `-s`/`-e` range and an explicit `-g` are the difference between a scan that finishes in a minute and one that sweeps an entire band. Not platform-specific; the same behaviour is there on any system.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andrepuschmann:
That looks good. Thanks @AndreiGosman

andrepuschmann:
@AndreiGosman are you happy to fill and sign the CLA form ?https://docs.srsran.com/projects/4g/en/latest/general/source/3_contributions.html

AndreiGosman:
> @AndreiGosman are you happy to fill and sign the CLA form ?https://docs.srsran.com/projects/4g/en/latest/general/source/3_contributions.html

Hello. Sure, form is signed and email sent.

andrepuschmann:
Thanks - the patch has been cherry-picked for internal testing and will make it into the next release.

